PR #122, opened 2026-08-02
chore: adopt the shared skill template

CybotTM:
Brings `.github/` in line with `netresearch/.github/templates/skill` and adds `.github/template.yaml`, which is what makes the template drift check apply to this repo from now on.

## Correction to the original description

The first version of this description claimed `release.yml` lacked the `id-token: write` and `attestations: write` scopes. **That was wrong** — all 17 repos in this rollout already had them. The claim came from reading a truncated `diff` and seeing only the template side of a changed hunk.

What `release.yml` actually gains is the top-level `permissions: {}` block and clearer comments. The scopes are unchanged.

## What this actually does

- adds `.github/template.yaml`, so `check-template-drift` governs this repo from now on
- adds the top-level `permissions: {}` deny-by-default block to the shared workflows that lacked it
- brings every shared workflow to the current template revision, including the scanner wording corrected in netresearch/.github#334
- adds the template files this repo did not carry

## Why overwriting is safe

Measured across all 17 skill repos in the same position before touching any of them: `release.yml` and `auto-merge-deps.yml` exist in **exactly one** revision each, the remaining files in two or three. No repo carries a local customisation — it is a single generation gap.

The one shape that would be unsafe is a repo hosting its own reusable workflow, since the template ships the **caller** form: adopting it there replaces the reusable with a stub that `uses:` itself, which is what broke every skill release org-wide on 2026-06-18. The rollout script refuses any repo containing `workflow_call:`, and none of the 17 does.

The commit message lists the files added and replaced for this specific repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,52 @@
+# Managed by netresearch/.github/templates/skill/
+#
+# Declares only the ecosystems every skill consumer is guaranteed to have:
+# github-actions (CI workflows) and composer (every skill repo ships a
+# composer.json for split-licensing / Packagist distribution).
+#
+# npm and devcontainers are OPT-IN: a repo that actually has package.json
+# fixtures (e.g. skills/<name>/references/examples/**) or a devcontainer adds
+# the block below to its own dependabot.yml and lists `.github/dependabot.yml`
+# under `intentional-drift:` in .github/template.yaml so the template sync
+# stops managing the file. Declaring an ecosystem without its manifest makes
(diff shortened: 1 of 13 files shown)


CybotTM:
Correction: the original description claimed `release.yml` lacked the `id-token`/`attestations` scopes. It did not — all 17 repos already had them, and the claim came from reading a truncated diff. The description has been corrected. What the file actually gains is the top-level `permissions: {}` block and clearer comments; the scopes are unchanged. The adoption itself stands: template governance, the deny-by-default block, and the current template revision.

PR #139, opened 2026-08-05
feat(pr-status): tell a queued check apart from a running one

CybotTM:
GitHub reports `queued` and `in_progress` as different check states. `pr-status.sh` collapsed both into "pending" and answered *required check(s) still running* — which reads as CI being slow when in truth no runner has picked anything up. The two situations have opposite answers.

## Why the distinction decides something

A merge queue drops an entry whose required check does not report within `check_response_timeout_minutes`, and that clock covers the wait for a **runner**, not just the run. `references/pull-request-workflow.md` already carried the warning — *"Do not enqueue into a busy runner pool"*, including the detail that a dropped entry does not cancel its runs so each retry makes it worse. But the tool an operator actually follows had no way to tell a busy pool from a slow one, so the warning could not be acted on.

Observed today on [netresearch/t3x-nr-llm#594](https://github.com/netresearch/t3x-nr-llm/pull/594): both merge-queue entries had `CI` at `status=queued`, never started, and the entry was discarded having executed nothing. Meanwhile 44 runs sat queued across 14 repositories with **zero** running. `pr-status` reported "still running" throughout.

## The change

- `QUEUED` is classified separately from `IN_PROGRESS`; `pending` downstream keeps its meaning of "not finished".
- New verdict **`await-capacity`** when a *required* check is queued while nothing at all is running, naming the checks and the counts.
- The checks line now reads `N pending (X running, Y queued)`.
- `checks.queued`, `checks.running` and `checks.queued_required` are exposed in `--json`.

## Deliberately not included: the limit

I looked for it rather than assuming:

```
$ gh api /orgs/netresearch/actions/hosted-runners/limits
{"message":"GitHub hosted runners are not supported for this organization","status":"404"}
$ gh api /orgs/netresearch --jq '.plan.name'
free
```

The endpoint covers larger runners only, and the per-plan ceiling for standard runners exists in documentation, not in the API. So the check observes state — required checks queued with zero running is sufficient on its own — instead of comparing against a number it cannot read.

## Verification

`shellcheck` and `bash -n` clean. Behaviour unchanged on three live PRs (`triage-ci`, `fix-ci`, and the merge-queue branch on #594 which still takes precedence, as it should).

The new branch could **not** be exercised against live data: the backlog drained while this was being written, so no PR currently has a queued required check. The classification and the branch condition are shown on synthetic input instead, which is weaker evidence and worth saying plainly:

```
$ jq -nc '[{status:"QUEUED"},{status:"IN_PROGRESS"},{status:"COMPLETED",conclusion:"SUCCESS"}] | map(...)'
[{"state":"QUEUED"},{"state":"PENDING"},{"state":"PASS"}]

# required queued, nothing running
await-capacity — required check(s) queued and not started: ci / CI — 6 queued, 0 running
# something running
wait — required check(s) still running
```

### DIFF
--- a/evals/evals.json
+++ b/evals/evals.json
@@ -150,6 +150,12 @@
       "name": "already-enqueued-pr-reads-as-clean",
       "prompt": "I queued my PR for merge ten minutes ago. The status check now says mergeState CLEAN, every check green, no open threads, NEXT: merge. Should I run the merge again?",
       "expected_output": "Should say no and check the PR's own queue entry first: mergeStateStatus describes the branch, so an enqueued PR keeps reading CLEAN and looks merge-ready; mergeQueueEntry (position, state, estimatedTimeToMerge) is the field that distinguishes 'ready to enqueue' from 'already in line', and a repo-level merge_queue ruleset does not answer it; should wait for the queue to merge it and note that enqueueing again only restarts its checks"
+    },
+    {
+      "id": 26,
+      "name": "required-check-queued-not-running",
+      "prompt": "My PR has been waiting on CI for forty minutes and the status tool says the required checks are still running. Should I enqueue it for merge so it goes in as soon as they finish?",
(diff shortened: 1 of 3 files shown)


CybotTM:
siehe oben

PR #145, opened 2026-08-06
fix(hooks): read the PreToolUse payload, add three forge gates

CybotTM:
`scripts/validate_git_command.py` has never fired for anyone who installs this skill. It reads a top-level `"command"` key, but the PreToolUse payload is `{"tool_name": ..., "tool_input": {"command": ...}}`. That JSON parses cleanly, so the `JSONDecodeError` fallback does not rescue it either — `command` comes out empty and the function returns before any check runs.

```
$ printf '%s' '{"tool_name":"Bash","tool_input":{"command":"git commit -m \"stuff\""}}' | python3 scripts/validate_git_command.py
                                     # (nothing)
$ printf '%s' '{"command":"git commit -m \"stuff\""}' | python3 scripts/validate_git_command.py
<system-reminder> … Commit message doesn't follow Conventional Commits format …
```

With `read_command()` accepting both shapes the existing advisory checks work as intended. That fix is the reason for the rest of the PR: three gates that belong to this skill had been living in a machine-local hook, invisible to anyone installing git-workflow. One of them recommends this repo's own `scripts/pr-status.sh`, so the skill was shipping the tool without the gate that points at it.

**What the three gates catch**

| Gate | Failure it prevents |
|---|---|
| Hard-wrapped forge body (3+ prose lines in the 55–85 column band) | Breaks read ragged in the web UI, break on mobile, and survive verbatim in release notes where markdown does not reflow |
| `repos/O/R/pulls/comments/{id}/replies` without the PR number | GitHub answers 404 and posts nothing — the reply looks sent |
| Sleep-loop over `gh pr view/checks/status` | Waits for the one outcome it was told about and sleeps through every other actionable event, including reviews |

They **deny** rather than warn: each describes an action that silently does the wrong thing, as opposed to one that merely reads badly. They run before the advisory checks, since a denied command never executes and warning about its commit-message style would be noise.

**Deliberate limits.** The reply gate only matches the `/replies` subresource and only in a segment that actually invokes `gh`/`curl` — `pulls/comments/{id}` is a legitimate read endpoint, and matching the path anywhere would block writing about it in an `echo`. The poll gate exempts `--watch`. The hard-wrap counter skips tables, lists, quotes, headings and fenced code, and ignores a lone short line, so a real one-line paragraph does not trip it.

A fourth gate stays machine-local on purpose: it resolves commit SHAs against the forge to catch hashes written from memory, which needs a network call and would not fit the 2 s hook timeout in `hooks/hooks.json`.

**Verification**

```
payload now read (advisory)     → <system-reminder> … Conventional Commits …
reply path without PR number    → deny
hand-rolled PR poll             → deny
pr-status.sh --watch            → passes
hard-wrapped --body (4 lines)   → deny
single-line --body              → passes
ls -la                          → passes
```

Ruff format and ruff check pass via pre-commit.


CybotTM:
## Review by hand — Copilot could not run

`copilot-pull-request-reviewer[bot]` answered: *"unable to review this pull request because the user who requested the review has reached their quota limit."* The standing rule is that a missing bot review gets done by hand, so this is a review of my own change, written to find fault with it. Two findings were serious enough to fix in this PR rather than file; the rest are stated as limitations.

### Fixed while reviewing

**`--body-file` on a pipe hung the hook.** The hard-wrap gate opened whatever path the flag named. `gh pr create --body-file <(generate-body)` is ordinary usage and hands over `/dev/fd/N`; opening it here waits for a writer this process cannot see. Measured before the fix: no return within 4 s against a fifo. A hook that hangs is worse than one that misses a finding — `hooks.json` allows it 2 s and every Bash call goes through it. Now non-regular paths are skipped and regular ones read to a 256 KiB cap. `a0bff80`

**A command prefix bypassed the reply gate.** `env FOO=1 gh api …/replies` and `sudo gh api …/replies` both passed, because the pattern anchored on `gh`/`curl` at the very start of the segment. Measured, both. The anchor cannot simply go — matching the reply path anywhere would block writing *about* it in an echo, which is why it was anchored originally — so it now skips leading `VAR=value` assignments and the usual wrappers first. `f02808c`

**No tests existed for this script at all**, which pairs badly with gates that deny. Sixteen cases now, and the useful half is the near-misses that must *not* fire: reading a single comment (same path prefix, no `/replies`), the reply path quoted inside an `echo`, a lone `gh pr view` that is not a poll, `pr-status.sh --watch`, a one-line body.

### Limitations I am not fixing here

**`read_command` keeps the old top-level `"command"` fallback.** For Claude Code payloads it is dead code. I kept it so the script stays callable directly, and because removing it would silently change behaviour for any other caller — but it is unexercised in practice and worth deleting once nothing else feeds this script.

**The 55–85 column band is a heuristic.** A paragraph whose lines genuinely fall in that range, three times consecutively, is denied. The `>= 3` threshold and the skipping of tables/lists/quotes/code keep it quiet in practice, but this is pattern-matching on prose and it will be wrong sometimes. The remedy for a false positive is obvious to the operator (unwrap it), which is why I consider the trade acceptable.

**`isValid`-style precision is missing on the poll gate.** `POLLS_PR` matches `pr-status.sh` itself, so a loop *around* `pr-status.sh` without `--watch` is denied. That is intended, but it means the gate assumes the script's own `--watch` is the only sanctioned wait — a fair assumption inside this skill, less so if someone wraps it deliberately.

**These three gates now exist in two places.** They still run in the machine-local hook they came from. Until that copy is stripped, a matching command gets two denies; the first wins, so the effect is cosmetic. I am deliberately not removing them before this merges — a window with *no* gate is the worse failure, and this repo has just demonstrated how long a silently-inactive hook can go unnoticed.

### What I checked and found sound

- The nested-payload fix is the reason the rest is worth anything: every advisory check in this file has been unreachable since it was written. The first test case pins exactly that.
- Deny runs before the advisory warnings. A denied command never executes, so commenting on its commit-message style would be noise.
- `reply_path_without_pr` splits on `&&`, `||`, `;`, `|`, `&` and newline, so a call in a later segment is still seen — covered by a case.
- The ported `hard_wrapped` counter is unchanged from the version that has been in service since 2026-08-02; I did not take the opportunity to "improve" it.
- `ruff@0.16.0 check` and `format --check` clean against the pinned CI version, after the first push failed on `FURB167`/`PIE810` that the local pre-commit did not raise.

### Evidence status

Every deny and pass above is a test case in `tests/test_validate_git_command.py`, 16 of 16 green. The two fixed defects were measured failing first — the fifo probe timed out at 4 s, and both prefix forms returned no deny — so the tests are known to sit in the position where the defect shows.

PR #166, opened 2026-08-10
feat(git-workflow): gate writes in the reference worktree, and three merge-gate fixes

CybotTM:
## Summary

Four findings from a `/retro` sweep of a session that took eight PRs through the merge queue. One new gate plus its recipe, and three corrections to guidance that was wrong or missing at the point it was needed.

## Came from

`/retro` session on 2026-08-10, project `t3x-nr-llm`.

### 1. Recipe 7 + `scripts/reference-worktree-gate.py` (C6 — written rule violated repeatedly)

- **Symptom:** four Bash calls in one session committed inside `~/projects/<project>/main`, producing a merge commit on the local `main`. Found three tool calls later, by accident.
- **Cause:** none of them said where they ran. Each inherited the working directory of an earlier call. `references/advanced-git.md` already states *"`main/` is reference only … Never commit, edit, or develop directly in it"*, and so did the operator's always-loaded instructions — prose cannot observe an inherited working directory, but a `PreToolUse` hook reads it out of the payload.
- **Required behavior:** deny the write; keep `git -C <path>` and `cd <path> &&` allowed, since both name where the write lands.
- **Verification:** `scripts/test_reference_worktree_gate.py` — 12 cases over a throwaway layout, including the exact command from the session and an end-to-end pass through the hook payload.

### 2. `new_duplicated_lines_density` is not always a re-attribution metric

- **Symptom:** the sole failing Sonar condition was `new_duplicated_lines_density` with 0 open issues — which this file described as grounds for merge-despite. It was measuring 191 lines of real copy-paste across three newly *added* sibling classes.
- **Cause:** the exemption was written for PRs that *move or touch* existing lines. It reads as unconditional, and `api/issues/search` cannot separate the cases — duplication is a measure, not an issue, so it returns 0 either way and the "issues are 0" half is satisfied in both.
- **Required behavior:** check which case you are in; ask `component_tree` which files carry the new duplicated lines. Added files → dedupe, no exemption.
- **Verification:** the query was run against a real response, which is how the two shapes now documented were found (`new_*` values live under `periods[0].value`, and the response lists directories as well as files).

### 3. A streaming watcher must emit on the transition

- **Symptom:** three identical "checks complete" notifications for one PR.
- **Cause:** the emit condition was a **state**, true on every subsequent poll. The `for`-loop skeleton in this file `exit`s and so never shows this; a `Monitor`-style streaming loop does.
- **Required behavior:** latch the last emitted message; one latch per PR when watching several.
- **Verification:** the snippet was run against a stubbed `gh` — three polls in the same state produce one line.

### 4. Serial merges duplicate append-only sections

- **Symptom:** ten `[Unreleased]` CHANGELOG bullets standing three and four times over; two PRs carried it to `main` before it was noticed.
- **Cause:** a resolver whose premise is *"ours holds only my additions"*. True for the first merge of a series, false for every later one — by then "ours" already contains what the previous merge brought in. No conflict markers, plausible diff.
- **Required behavior:** rebuild the section from *theirs* plus your own block instead of merging it, and assert heading count / single occurrence / no repeated first lines.
- **Verification:** those three assertions are what eventually caught it.

## Test plan

- [x] `scripts/test_reference_worktree_gate.py` — 12/12
- [x] every script in `tests/` — 8/8
- [x] `Build/Scripts/validate-skill.sh` — 0 errors, 0 warnings
- [x] pre-commit (markdownlint, ruff, ruff-format) clean; tests re-run after `ruff format` touched both new files
- [x] every snippet added here executed before committing (jq against a real SonarCloud response, latch against a stubbed `gh`, `bash -n` on the fragment)

### DIFF
--- a/scripts/reference-worktree-gate.py
+++ b/scripts/reference-worktree-gate.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+"""PreToolUse gate: refuse a git write inside a reference `main/` worktree.
+
+In the bare-repo layout (`<project>/.bare` + one directory per branch), `main/`
+exists to be read, fetched from, and used as the base for new worktrees.
+`references/advanced-git.md` states that and so do most AGENTS.md files that
+adopt the layout. The rule still gets broken, because breaking it needs no
+intent: one Bash call omits its `cd`, inherits the working directory of an
+earlier call, and the commit lands on the shared branch locally — where a later
+push or a `jj git export` moves it under every sibling worktree.
+
+This denies the write. It does not deny naming the target explicitly:
+`git -C <path> commit` and `cd <path> && git commit` both pass, because both say
+where the write lands.
+
+Fails open. Any error, an unreadable payload, or a directory outside the layout
+allows the command — a gate that blocks on its own failure is worse than the
+friction it prevents.
+"""
+
+import json
+import os
+import re
+import sys
+
+# `<root>/<project>/main` (or `master`) — the reference checkout of the layout.
+# Set GIT_WORKTREE_ROOTS to a colon-separated list to match other roots.
+DEFAULT_ROOTS = ("~/projects", "~/p")
+
+# Subcommands that leave a commit, an index entry, or rewritten history behind.
+# `add` is included deliberately: the damage this gate exists to stop starts
+# with a stray `git add -A` staging a directory nobody meant to touch.
+GIT_WRITE = re.compile(
+    r"\bgit\s+(?:commit|add|merge|rebase|cherry-pick|revert|am|stash)\b"
+    r"|\bgit\s+reset\s+--hard\b"
+)
+# `git -C <path>` moves the operation out of the working directory, so the
+# working directory no longer decides where the write lands. That is the
+# explicit form this gate asks for, so it is never blocked.
+GIT_DASH_C = re.compile(r"\bgit\s+-C\s+\S+")
+
+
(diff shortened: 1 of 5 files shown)


CybotTM:
Copilot failed here too — the COMMENTED row carries an error in its body rather than a review. Reviewed it myself; the parts that needed proof were executed rather than reasoned about:

- **The gate** reproduces the friction it was written for. `scripts/test_reference_worktree_gate.py` runs 12 cases over a throwaway layout including the exact command from the session (`git add -A && git commit -S --no-edit -q` in `main/`), and the six negative cases that carry the rule — `git -C <path>`, a leading `cd`, reads, `fetch`, `worktree add`, and a same-named directory without a `.bare` sibling.
- **The jq** was run against a real `component_tree` response. That is how the two documented failure shapes were found: a `new_*` metric carries its value under `periods[0].value`, and the response lists directories alongside files. The first version I wrote returned nothing and would have shipped as a plausible-looking recipe.
- **The latch snippet** was run against a stubbed `gh`: three polls in the same state produce one line, then the terminal state.

Gates: 8/8 repo test scripts, `validate-skill.sh` 0/0, pre-commit clean, and `ruff check .` / `ruff format --check .` at the versions CI pins. The last commit exists because the pre-commit ruff alias runs a narrower rule set than CI does — EXE001 and PLW1510 surfaced only in CI.

No SKILL.md change, so the 500-word cap is untouched.

PR #167, opened 2026-08-10
fix(git-workflow): let the reference worktree refresh itself

CybotTM:
## Summary

The gate added in #166 over-blocked: it denied `git merge --ff-only origin/main` inside a `main/` worktree, which is the one write that directory exists for.

## Came from

The gate blocked its own author's `git reset --hard origin/main` in `main/` **minutes after installation** — while verifying that an unrelated PR had landed. The block was correct in kind, and the question it forced ("is this actually wrong?") is what separated the two cases:

- A reference checkout has to be **current** to serve as the base for new worktrees. Refreshing it is the directory's purpose, and `merge --ff-only` / `pull --ff-only` can neither create a commit nor discard work → **exempt**.
- `reset --hard` reaches the same state *by discarding*, which is exactly the case worth stopping → **still refused**.
- A plain `merge` lands a merge commit there → **still refused**.

## Change

- `FAST_FORWARD_ONLY` exemption in `scripts/reference-worktree-gate.py`, beside the existing `git -C` one.
- The denial message names the allowed refresh, so the reader is pointed at the right command rather than only away from the wrong one.
- Recipe 7 states the exemption and why `reset --hard` is not part of it.
- Four cases in `scripts/test_reference_worktree_gate.py` pin both directions.

## Test plan

- [x] `scripts/test_reference_worktree_gate.py` — 16/16, including the two now-allowed refresh forms and the two still-refused ones
- [x] every script in `tests/` passes
- [x] `validate-skill.sh` — 0 errors
- [x] `ruff check .` and `ruff format --check .` at the versions CI pins

### DIFF
--- a/scripts/reference-worktree-gate.py
+++ b/scripts/reference-worktree-gate.py
@@ -38,6 +38,12 @@ GIT_WRITE = re.compile(
 # working directory no longer decides where the write lands. That is the
 # explicit form this gate asks for, so it is never blocked.
 GIT_DASH_C = re.compile(r"\bgit\s+-C\s+\S+")
+# Bringing the reference checkout up to date is what the directory is FOR — it
+# has to be current to serve as the base for new worktrees. A fast-forward
+# cannot create a commit or discard work, so `merge --ff-only` / `pull --ff-only`
+# are the sanctioned refresh and stay allowed. `reset --hard` is not: it reaches
+# the same state by discarding, which is the case worth stopping.
+FAST_FORWARD_ONLY = re.compile(r"\bgit\s+(?:merge|pull)\b[^|&;]*--ff-only\b")
 
 
 def roots() -> list[str]:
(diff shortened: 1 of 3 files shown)


CybotTM:
Copilot failed again — the COMMENTED row carries an error in its body rather than a review, as on #166. Reviewed it myself; the change is small and the risk is entirely in the regex boundary, so that is what I checked.

`FAST_FORWARD_ONLY` is `\bgit\s+(?:merge|pull)\b[^|&;]*--ff-only\b`. The `[^|&;]*` matters: it keeps the flag bound to the same shell statement, so `git merge origin/main && echo --ff-only` does not exempt the merge. Verified against the live hook on this machine:

| command in `main/` | verdict |
|---|---|
| `git merge --ff-only origin/main` | allowed |
| `git pull --ff-only` | allowed |
| `git fetch origin -q && git merge --ff-only origin/main` | allowed |
| `git merge origin/main` | denied |
| `git reset --hard origin/main -q` | denied |
| `git add -A && git commit -q -m x` | denied |

The asymmetry is the point and is stated in the recipe: `--ff-only` cannot create a commit or discard work, `reset --hard` reaches the same end state by discarding, and a plain `merge` lands a merge commit in the reference checkout.

Gates: `scripts/test_reference_worktree_gate.py` 16/16, every script in `tests/` passes, `validate-skill.sh` 0 errors, `ruff check` / `ruff format --check` at the pinned versions.

PR #171, opened 2026-08-12
fix(pr-status): a cancelled run is not a failed one, and CI readiness needs its own field

CybotTM:
Two readings that sent an operator after the wrong thing during a 19-PR merge series on 2026-08-12.

**CANCELLED counted as FAIL.** Every conclusion that is not SUCCESS, SKIPPED or NEUTRAL fell into `else "FAIL"`. When a push supersedes a run, its jobs conclude CANCELLED and those check-runs stay on the earlier commit forever, so a pull request GitHub itself called CLEAN read as `2 fail` with `NEXT: fix-ci`. The rows were doubly opaque because the workflow's `name:` is an unevaluated expression, so they showed up as `e2e / matrix.typo3 != '' && …`. An hour went into a run that had been cancelled on purpose.

CANCELLED is now its own state and split by whether the context reported again:

- reported again under the same name -> `stale`, dropped from the counts;
- never reported again -> still shuts the gate, but listed as `cancelled — re-run, do not debug` rather than folded into `failing`.

Merge semantics are unchanged: a genuinely unmet context still counts.

**No field said whether CI had started.** The review branches of the NEXT ladder sit above every CI branch, so on a repo with the copilot_code_review ruleset NEXT answers `request-review` from the second a commit is pushed and never mentions CI. The checks counts do not fill the gap either: right after a push `0 pending` is true because one context has registered and the other ninety have not. `ci-cd-integration.md` already warned about that trap in prose, but there was nothing machine-readable to gate on, so an automated caller enqueued a pull request whose CI had not started and the merge queue refused it.

`checks_settled` is now in the JSON — true only when nothing is pending and every required context has reported at least once — and the request-review answer carries the CI state in its `why`. The ladder order is unchanged: the review really is the blocking gate there.

Also documents how to read a failed job's output at all. `gh api .../actions/jobs/<id>/logs` and `gh run view --job <id> --log` return the runner's preamble and nothing else (one of them an empty body, exit 0); the run's log archive has the step output. That was the difference between three unexplained merge-queue ejections and the one line naming the cause.

Verified against the pull requests that misreported: #728 goes from `2 fail` to `1 fail` plus an explicit stale/cancelled breakdown, and `checks_settled` is false there (four required contexts never reported on that head) while true on a sibling that settled normally.

### DIFF
--- a/skills/git-workflow/references/ci-cd-integration.md
+++ b/skills/git-workflow/references/ci-cd-integration.md
@@ -44,6 +44,37 @@ Because the exit code cannot distinguish "watched and passed" from "never
 found the run", assert the run exists before watching it, or re-read the
 check's state afterwards rather than trusting the watcher's return.
 
+### The step output of a failed job comes from the RUN's log archive
+
+`gh api repos/$R/actions/jobs/$JOB/logs` and `gh run view --job $JOB --log`
+routinely hand back only the runner's own preamble — image provisioning,
+hardening-agent chatter, `Cleaning up orphan processes` — with the failing
+step's output absent entirely. One of them can also return an empty body and
+exit 0. Filtering that noise then produces nothing and reads as "the log says
+nothing about why it failed", which sends you diagnosing the wrong layer.
+
+Download the **run's** archive instead. It contains one text file per job, with
+the real step output:
+
+```bash
+gh api "repos/$R/actions/runs/$RUN/logs" > /tmp/logs.zip
+unzip -q -o /tmp/logs.zip -d /tmp/logs
+grep -rn '##\[error\]' /tmp/logs | grep -viE 'armour|agentservice|pam_unix'
+```
+
+On 2026-08-12 this was the difference between three unexplained merge-queue
+ejections and one line — `[ERROR] Undefined constant …PHPUnitSetList::PHPUNIT_110`
+— that named the cause outright. The job-level calls had been tried first and
+showed nothing but setup.
+
+Note the archive is per *run*, so for a queue ejection you need the
+`gh-readonly-queue/*` run, not the pull request's own:
+
+```bash
+gh run list --repo "$R" --limit 20 --json headBranch,databaseId,conclusion \
+  --jq '.[] | select(.headBranch | startswith("gh-readonly-queue"))'
+```
+
 Hand-rolled `gh pr checks | jq` poll loops re-derive those semantics from
 undocumented field shapes (a running check's `conclusion` may be `""`, `null`,
 or absent) and are a recurring source of bugs. The sharpest one: a poll run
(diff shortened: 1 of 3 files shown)


CybotTM:
## Review note — Copilot out of quota, reviewed by the author

`copilot_code_review` is active and Copilot answered on both heads with *"unable to review this pull request because the user who requested the review has reached their quota limit"*. That is a quota ceiling, not an outage, so re-requesting cannot produce a review — the skill's own fallback applies: review it yourself and say so.

The diff was read against its final head. It found **three defects in the first commit**, all fixed in `4c079dd`:

- **The printed counts stopped adding up.** `total` stayed the raw row count while the four states beside it excluded stale rows, so a pull request rendered as `86 pass, 1 fail, 0 pending, 12 skip (of 100)` — 99 accounted for, nothing saying where the hundredth went.
- **A multi-line check-run name broke the summary.** Where a workflow's `name:` is an unevaluated expression the name arrives with newlines in it, and the one-line `cancelled` entry split into fragments mid-report.
- **`checks_settled` could read true on a set it never verified.** It derives partly from `undispatched`, which is empty both when every required context reported and when the rules fetch failed — the exact conflation this script's header calls out as the one direction it must not get wrong. It is now `null` when `rules_fetched` is false.

A fourth was caught by `bash -n` before it was committed: the first attempt at the second fix put an apostrophe into a comment inside the jq program, which ends the single-quoted shell string. The file warns about that twice; I did it anyway.

**Verified against the pull requests that misreported**, not just re-read:

| | shipped | this branch |
|---|---|---|
| netresearch/t3x-nr-llm#728 | `86 pass, 2 fail` — `NEXT: fix-ci` | `86 pass, 1 fail (of 99)` + `stale: 1 … ignored` + `cancelled: … re-run, do not debug` |
| netresearch/t3x-nr-llm#745 | — | `checks_settled: true`, `stale: 0`, counts sum to total |

`checks_settled` is `false` on #728 for a real reason (four required contexts never reported on that head, the queue branch having satisfied them) and `true` on #745 — so the flag distinguishes the two rather than firing on everything.

Stated plainly because it bears on how much this is worth: this is the author reviewing his own change to shared tooling. It is not a substitute for a second pair of eyes, and it is what was available.

CybotTM:
Rebased onto the moved `main` (`--force-with-lease`, commits re-signed with `--exec "git commit --amend --no-edit -S"`; both carry a valid signature and the DCO trailer).

The self-review above was written against the pre-rebase head, so its SHAs no longer exist. The review still stands: the branch's own delta is **byte-identical** before and after the replay — verified as `diff <(git diff <old-base>..<old-tip>) <(git diff origin/main...HEAD)`, empty — and the rebase resolved no conflicts. The behavioural check was re-run on the new head and gives the same answer on the pull request that originally misreported.

PR #173, opened 2026-08-12
docs(merge-gate-watcher): note merge-queue strategy flag and quota --watch timing

CybotTM:
Two field notes added to `references/merge-gate-watcher.md`, from a session driving several PRs to merge on merge-queue repos while the Copilot reviewer was out of quota. Both fill gaps the doc did not cover; no restructuring.

- **Merge-queue strategy flag.** `gh pr merge $PR --merge` (or `--squash`/`--rebase`) on a merge-queue repo prints `! The merge strategy for main is set by the merge queue` and ignores the flag — but it still enqueues the PR, so the warning is a notice, not a failure. Call it without a strategy flag there. (The watcher skeleton uses `--merge`, which is why the clarification is worth stating explicitly.)
- **Quota `--watch` timing.** Once the review quota is exhausted, `pr-status.sh --watch` returns on the `request-review` event immediately, even while CI is still running — the review-state event is independent of check completion, so it fires before the gate can be `CLEAN`. After deciding to proceed unreviewed, re-arm until `mergeState=CLEAN` before merging; only `CLEAN` passes the gate (`BLOCKED`/`UNSTABLE` do not). Added to the existing bot-quota section that already covers "a bot review can be a failure notice".

Captured via `/retro`. Docs-only, 4 lines.

### DIFF
--- a/skills/git-workflow/references/merge-gate-watcher.md
+++ b/skills/git-workflow/references/merge-gate-watcher.md
@@ -100,6 +100,8 @@ done
 
 Pitfalls baked in: `grep -c` exits 1 on zero matches (`|| true`); decide hard-fail only at `PENDING -eq 0` (codecov posts transient FAILURE mid-run); never count a check class you did not explicitly list.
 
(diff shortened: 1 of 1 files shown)


CybotTM:
Copilot review returned a quota-limit error (no findings). This is a 4-line docs-only change to references/merge-gate-watcher.md — two field notes appended to existing sections, no restructuring — reviewed by the author. Required checks green.

PR #174, opened 2026-08-12
docs(pull-request-workflow): pr-status.sh is GitHub-only, name the GitLab path

CybotTM:
Nothing in the script's name or in the section that introduces it says it cannot answer for a GitLab merge request. It speaks GitHub GraphQL, so `-R git.netresearch.de/group/project` is accepted at the command line and then fails with a bare `pr-status: GraphQL query failed` — and `--watch` produces no output at all and simply sits there. A GitLab MR ended up behind such a watcher for 560 s before the mismatch was noticed, which is exactly the "waiting on something that can never fire" failure the surrounding guidance exists to prevent.

Added a short scope statement where the script is introduced, plus the two `glab api` calls that answer the same question for GitLab: `detailed_merge_status` on the merge request — which, unlike `mergeStateStatus`, does name the reason — and the `/discussions` endpoint for unresolved threads.

Documentation only, no script change. `markdownlint-cli2` clean.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -28,6 +28,23 @@ Measured on a 40-PR rollout that did not have it: **183 of 370 shell calls
 were PR-status probing**, the rulesets endpoint was queried exactly once, and
 `copilot_code_review` blocked four merges by surprise.
 
+**GitHub only.** The script speaks GitHub GraphQL, so it cannot answer for a
+GitLab merge request — `-R git.netresearch.de/group/project` is accepted at the
+command line and then fails with a bare `pr-status: GraphQL query failed`, and
+`--watch` produces nothing at all until it is killed. Nothing in the name says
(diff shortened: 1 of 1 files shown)


CybotTM:
Self-reviewed, because Copilot failed on both heads of this PR — three failed attempts across two PRs today, which is a quota ceiling rather than an outage that clears by asking again. Recording what was checked instead of merging unreviewed.

The review found one real defect, fixed in 10af26d. The snippet showed the URI encoding the project path needs on its first `glab` call and then used a literal `projects/...` placeholder on the second — copying the block gave one line that runs and one that does not. Both calls now share an encoded `$P`, and the block parses under `bash -n`.

Verified on the claims themselves rather than on plausibility: `pr-status: GraphQL query failed` and the silent `--watch` were both observed against a real GitLab MR in the session this came from, and `detailed_merge_status` returning a named reason (`mergeable`, `checking`) was read back from that same MR. `markdownlint-cli2` clean, all checks green, documentation only — no change to the script.

PR #176, opened 2026-08-13
docs(no-editorializing): name the pipe that breaks the line-wrapping rule

CybotTM:
The line-wrapping rule in `no-editorializing.md` was already here and already correct — it lists PR/MR descriptions, review comments, issue bodies and release notes as comment surfaces where a hard wrap renders as a `<br>`. This is not a gap in what the reference says.

It was still broken on **six of six** pull requests created in one session, by someone who had this reference loaded. Every body and comment the same person typed by hand in that session measured clean. So the difference was not knowledge — it was that the violating step does not look like writing prose at all:

```bash
git log -1 --format=%b > /tmp/body.md
gh pr create --body-file /tmp/body.md
```

A commit body is *correctly* hard-wrapped at ~72 columns, and `commit-conventions.md` says so two files over. Piping it into a description carries that wrapping across the boundary the section is about, and nothing in the command reads as wrong.

## What this adds

- The trap itself, named, with the command that causes it.
- A join-and-strip recipe that also removes `Signed-off-by`, which is noise in a description rather than a trailer anyone reads there.
- The caveat that the recipe is deliberately naive: it will flatten a fenced block or a list, so the output needs reading before it is posted. Verified by running it against a real commit body and against a list.
- A detector worth grepping for in an audit: a `Signed-off-by:` visible in a rendered PR description means that description came from a commit message, which means it is wrapped. That is how the six were found.
- A forward reference from `commit-conventions.md` at the 72-column rule, which is where the wrong idea starts.

## Why documentation and not a check

A mechanical gate would have to live in whatever creates the pull request, and `gh pr create --body-file` is not this skill's to wrap. The realistic guard is the fingerprint: one grep over `gh pr view --json body` finds every body that came from a commit message, and that is cheap enough to run in a sweep.

### DIFF
--- a/skills/git-workflow/references/commit-conventions.md
+++ b/skills/git-workflow/references/commit-conventions.md
@@ -93,7 +93,9 @@ EOF
 ```
 
 **Hard-wrap the commit *body* (~72 cols); don't hard-wrap prose whose wrapping
-you don't control.** The 72-column convention is for commit message bodies. For
+you don't control.** The 72-column convention is for commit message bodies —
+which is exactly why a commit body must not be piped into a PR description
+unchanged; see *The rule breaks at the pipe* in `no-editorializing.md`. For
 **PR/MR & issue descriptions, release notes, review comments, and chat**, write
 one line per paragraph (and per list item) and let the renderer soft-wrap — a
 hard break mid-paragraph there does nothing for the output. For **committed
--- a/skills/git-workflow/references/no-editorializing.md
+++ b/skills/git-workflow/references/no-editorializing.md
@@ -58,3 +58,45 @@ pipelines, and hard-wrapping prose is right in one and wrong in the other.
 Do not wrap a release body or PR/comment the way you wrap a `.md` file. Fix one
 already published with hard-wraps via `gh release edit <tag> --notes-file <file>`
 (release bodies stay editable) or by editing the PR/comment.
+
+### The rule breaks at the pipe, not at the keyboard
+
+Knowing the rule is not enough, because the way it gets violated does not feel
+like writing prose at all:
+
+```bash
+git log -1 --format=%b > /tmp/body.md
+gh pr create --body-file /tmp/body.md      # <-- wrapped, every time
+```
+
+A commit body is *correctly* hard-wrapped at ~72 columns — `commit-conventions.md`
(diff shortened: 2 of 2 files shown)


CybotTM:
## Review note — Copilot out of quota, reviewed by the author

Copilot answered on `35c6a110` with the quota error, so the skill's own fallback applies. This patch is documentation, so the review is mostly "are the claims in it true" — and it makes three checkable ones.

**"Six of six pull requests created that way were wrapped."** Measured across all 23 pull requests from the session, counting prose lines in the 58–92 column window outside fenced blocks: `t3x-nr-llm#748` 17/20, `#749` 17/19, `#750` 9/9, `#753` 25/31, `git-workflow-skill#171` 31/36, `github-release-skill#82` 12/16. Those six are exactly the ones whose body came from `git log --format=%b`.

**"Every body and comment typed by hand in the same session was clean."** The other 17 pull request bodies, written directly as bodies, scored 0–2 out of 10–16. Issue bodies and every review comment scored 0. Seven lines flagged across the comments on this repo's #171 and on `github-release-skill#82` turned out to be a heading, a table row, and complete sentences — inspected individually, none a wrapped fragment.

**"A `^Signed-off-by:` in a rendered body is the fingerprint."** Run in both directions: 1 on each of the four wrapped originals, 0 on each hand-written body and comment. It detects, and it does not false-positive on the clean set.

Two things the review changed before this was posted:

- The `awk` recipe originally left doubled spaces where a list item's continuation was joined (`- Punkt eins   mit Fortsetzung`). Added `gsub(/  +/," ")` and re-ran it against a list and a real commit body.
- My first pass at counting trailers used an unanchored `grep -c 'Signed-off-by'`, which reported 2 on this pull request's own body — both were the prose *describing* the detector. The version documented here is anchored to `^`, which is what makes it a detector rather than a word search. Worth stating because the unanchored form is the one that comes to mind first.

**On not shipping a mechanical gate instead.** The natural place for one is whatever calls `gh pr create --body-file`, which is not this skill's code. The fingerprint grep is the realistic guard: one call per pull request over `gh pr view --json body`, cheap enough for a sweep, and it is how the six were found in the first place.

This is the author documenting his own mistake, in a reference he had loaded at the time and did not apply. The rule was already correct and already complete — what is added is the one path that defeats it.

PR #177, opened 2026-08-13
docs(merge-gate-watcher): ask git whether a PR merged, and reach REST when GraphQL is dead

CybotTM:
Two additions to `merge-gate-watcher.md`, both from burning the GraphQL budget twice in one session while doing nothing unusual.

## "Has it merged yet?" needs no API at all

The existing "Watcher cost" section makes a watcher cheaper — one per subject, REST over GraphQL, 180 s. It does not say that the commonest watcher question is free: once a PR is queued, the only thing left to learn is whether its head landed on the base, and `git merge-base --is-ancestor` answers that from the local graph. It keeps working while both budgets are exhausted, which is exactly when a watcher is still running.

Added with the trap that bit me: **`git log --grep="#<pr>"` is not that test.** `--grep` searches the whole commit message, so an older commit whose body mentions the number matches. A watcher on PR #765 reported `MERGED` from its first tick because a months-old dependency bump mentioned that number; the PR was actually `CONFLICTING`. Same asymmetry as the section's own empty-result rule, inverted — an empty result is first a broken query, and a positive result from a text search is first a coincidence.

## REST still opens a PR when GraphQL is dead

`gh pr create` is GraphQL. With `graphql: 0/5000` and `core: 4700/5000` the whole `gh pr *` surface is unusable while REST is fine, and `gh api repos/$O/$R/pulls -X POST` takes the same arguments. Same for PR comments. Both used on 2026-08-13 to finish work that `gh pr create` refused.

Plus one flag trap found next to it: `gh api --paginate --slurp` is rejected together with `--jq`, so paginated check-runs need a file and then `jq`.

## Not added

The section already says to sleep to the reset in one background command rather than retrying into the limit. I violated that rule rather than found it missing, so it needs no more prose.

### DIFF
--- a/skills/git-workflow/references/merge-gate-watcher.md
+++ b/skills/git-workflow/references/merge-gate-watcher.md
@@ -294,6 +294,8 @@ git merge-base --is-ancestor "$HEAD_SHA" origin/main && echo MERGED
 
 `$HEAD_SHA` is the PR head you pushed, which you already know. This keeps working while both budgets are exhausted, which is exactly when a watcher is most likely to be running. Reserve `pr-status.sh` for the merge *gate* — checks, threads, reviews, the `NEXT` line — and use git for the merge *fact*.
 
+**Ancestry answers only where the merge preserves the commit.** `--merge` and `--rebase` do; **squash does not** — it writes one new commit with a new hash, so the original head is never an ancestor and the check reads "not merged" forever. In a squash-merge repo, accept one cheap REST call instead: `gh api repos/$R/pulls/$PR --jq .merged`. Know which strategy the repo allows before relying on ancestry — `pr-status.sh` prints it as `merge: methods=[…]`.
+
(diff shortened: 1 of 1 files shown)


CybotTM:
Self-review (Copilot has been quota-blocked all day across both repos, so this is the review of record — and re-requesting a fourth time is the exact "retry into a known-exhausted resource" this PR argues against).

**Two findings, both fixed here rather than after merge.**

*The false-positive example was unverified prose.* It said "a months-old dependency bump mentioned `#765` in its body" — I had not looked. Checked: the matched commit is [this January dependency bump](https://github.com/netresearch/t3x-nr-llm/commit/be14b02c) from **2026-01-12**, seven months before the watcher ran, and the `#765` in it is an issue number belonging to `actions/attest-build-provenance` — a *different repository*, embedded in the changelog Dependabot pastes into the body. A sharper trap than the one I described, so the text names it.

*The ancestry check silently fails under squash.* A squash merge writes one new commit with a new hash, so the original head is never an ancestor and the check reads "not merged" forever — a watcher built on this advice would hang in any squash-merge repo. Added the limit and the one-REST-call alternative. This is the finding I care about: the recommendation was correct for the repo I was standing in and wrong as a general rule.

Refutations that did not hold:

- *A fetch costs budget too.* It does not — the git protocol is neither the REST nor the GraphQL quota. That is the entire point of the section.
- *The REST fallback is speculative.* Used twice on 2026-08-13 to open netresearch/t3x-nr-llm#764 and #765 after `gh pr create` returned `GraphQL: API rate limit already exceeded`.
- *This duplicates the existing budget rules.* It does not restate them; the section says so explicitly and adds only what they omit — that the commonest watcher question needs no budget at all.

PR #178, opened 2026-08-13
docs(pull-request-workflow): a stacked PR loses its approvals when its base merges

CybotTM:
`pull-request-workflow.md` describes stacking nowhere, and the one behaviour that makes it surprising is undocumented: when the base PR merges, GitHub retargets the child to `main` **and dismisses every review on it**. The child reports `decision=REVIEW_REQUIRED` with an unchanged head SHA, green checks and no file moved — which reads like a bug rather than a base-change side effect.

It is a host gate, so the child cannot merge until an approval lands on that head again, and an automated approver only reruns on a *new* head. `gh api .../update-branch -X PUT` supplies one, at the cost of the full check matrix a second time.

Added next to `Handling Stale PRs`, with the two planning consequences: budget the second CI run when you choose to stack, and expect the child's tree to conflict after the base advances even though its diff is now correct — a shared `CHANGELOG.md` "Unreleased" section being the usual case.

Observed on netresearch/t3x-nr-llm#759, stacked on #758 for a documentation anchor that existed only on that branch.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -915,13 +915,15 @@ Stacking — PR B opened against PR A's branch so B can build on text or code th
 reviews : github-actions=DISMISSED   decision=REVIEW_REQUIRED
 ```
 
-Nothing about B changed. Its head SHA is the same, its checks are still green, no file moved. But `REVIEW_REQUIRED` is a host gate, not advice, so B cannot merge until an approval lands on that head again — and an automated approver only reruns on a new head. The recovery is to give it one:
+Nothing about B changed. Its head SHA is the same, its checks are still green, no file moved. But `REVIEW_REQUIRED` is a host gate, not advice, so B cannot merge until an approval lands on that head again — and an automated approver only reruns on a new head. The recovery is to give it one, with the plain (merging) form of the command in [Updating a PR branch without a local clone](#updating-a-pr-branch-without-a-local-clone--gh-pr-update-branch---rebase):
 
 ```bash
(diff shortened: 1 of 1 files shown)


CybotTM:
Self-review (Copilot quota-blocked all day; this is the review of record).

**One finding, fixed here.** The section reinvented `update-branch` as a raw `gh api … -X PUT` call while this very file already documents `gh pr update-branch` a few hundred lines below. Now it uses the documented command and cross-references that section — and says how the two cautions differ, because they are easy to conflate: there, `--rebase` force-updates and *can* reset approvals; here the approvals are **already** gone before you touch anything, dismissed by the retarget itself, and the update is what restores them.

Refutations attempted:

- *The dismissal is an org policy, not GitHub behaviour.* The repo's rulesets are `copilot_code_review, deletion, merge_queue, non_fast_forward, pull_request, required_signatures, required_status_checks` — none of which dismisses on base change. It is the base-change dismissal that GitHub applies to any review when a PR is retargeted.
- *The head SHA must have changed, so of course reviews dropped.* It did not: netresearch/t3x-nr-llm#759 reported the same head before and after the retarget, with `checks: 72 pass, 0 fail` unchanged, and only `github-actions=APPROVED` → `DISMISSED  decision=REVIEW_REQUIRED`.
- *Stacking is rare enough not to document.* It is the correct shape whenever the child has no anchor without the parent — here a documentation paragraph that existed only on the parent branch — and the surprise costs a full CI matrix each time.

The conflict warning in the last bullet is not speculative either: after the base merged, the child conflicted on the shared `CHANGELOG.md` "Unreleased" section, twice in the same session.

PR #180, opened 2026-08-13
feat(pr-status): make the --json contract discoverable and enforced

CybotTM:
`pr-status.sh --json` did not describe itself, so a caller writing a jq filter against it had to guess — and a wrong guess is silent. jq answers a missing key with `null` and says nothing, so a loop waiting for `.mergeStateStatus == "CLEAN"` (the GraphQL name `gh pr view` uses, not this script's `mergeState`) never fires and reads as "still running" forever.

That is not hypothetical: it happened to me on 2026-08-13 while watching two PRs of this repo. The loop ran to its iteration limit reporting `null` every tick, on PRs that had been `CLEAN` for ten minutes.

## What this adds

**A `--json contract` block in the header**, which is what `-h/--help` prints — the place I should have looked. It lists every top-level key, names the two traps (the field names are this script's, not GraphQL's; a wrong key is silent), and points at `--watch` first, because in my case the whole loop was a hand-rolled substitute for a flag that already exists.

**`--help` no longer truncates.** It printed a hard-coded line range, so the header could grow past it and the new block would have been invisible — the exact failure the block exists to prevent. It now prints the leading comment block, however long it gets.

**`tests/test_pr_status_json_contract.sh`** compares the documented list against what a stubbed run actually emits, in both directions, and asserts `--help` still shows it. A contract nobody checks rots at the first new field; `checks_settled` was added recently and would have been the first to go missing.

## Verification

Controlled failure in all three directions, not just a green run:

| injected fault | result |
|---|---|
| header lists `bogusField` | `FAIL … header lists keys --json does not emit: bogusField` |
| header drops `checks_settled` | `FAIL … --json emits keys the header does not list: checks_settled` |
| `--help` back to a fixed line range | `FAIL --help does not show the '--json contract' block` |

Restored: all three ok. Full suite: 8/8 shell tests pass, including `test_pr_status_errored_review.sh`, which drives the same script.

No behaviour change to `--json`, `--watch` or the prose rendering.

### DIFF
--- a/skills/git-workflow/scripts/pr-status.sh
+++ b/skills/git-workflow/scripts/pr-status.sh
@@ -24,6 +24,35 @@
 # soon as something can be worked on: a failing check, a new annotation, or all
 # required checks concluded. Waiting for `pending == 0` means learning nothing
 # until the slowest matrix job ends, long after the first failure was visible.
+#
+# --json contract
+#
+# One object, and its field names are THIS SCRIPT'S — not the GraphQL names
+# `gh pr view` uses. `mergeState`, not `mergeStateStatus`. `base`, not
+# `baseRefName`. Guessing costs more than reading: jq answers a missing key
+# with `null` and says nothing, so a loop waiting for
+# `.mergeStateStatus == "CLEAN"` never fires and reads as "still running"
+# forever. Top-level keys:
+#
+#   state mergeable mergeState draft number title repo author
+#   base head headOid
+#   checks checks_settled threads unresolved_threads
+#   reviewDecision reviews_on_head has_review_on_head
+#   has_copilot_review_on_head copilot_review_errored copilot_error_count
+#   requested_reviewers
+#   merge_methods auto_merge_allowed queue_active queue_entry
+#   rulesets rules_fetched required_contexts undispatched unsigned
+#   next
+#
+# `checks` is {total,pass,fail,pending,skip,...} and `next` is
+# {action,why,cmd} — the same two things the prose rendering leads with.
+#
+# Before writing a jq filter against any of these, consider whether --watch
+# already answers the question; it usually does, and a hand-rolled poll loop is
+# how the wrong field name gets guessed in the first place.
+#
+# tests/test_pr_status_json_contract.sh fails when this list and the emitted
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review (Copilot quota-blocked all day; this is the review of record).

This PR exists because I guessed a field name against this script's own `--json` and then reported a merge state from `null`. Reviewing it, the thing worth stating plainly: the header alone would have been the weaker half of the fix. A documented contract nobody checks rots at the first new field, and `checks_settled` — added recently — would have been the first to go missing. The test is the part that keeps it true.

**Verified by controlled failure, three ways**, because a green new test proves only that it runs:

| injected fault | result |
|---|---|
| header lists `bogusField` | `FAIL … header lists keys --json does not emit: bogusField` |
| header drops `checks_settled` | `FAIL … --json emits keys the header does not list: checks_settled` |
| `--help` back to a fixed line range | `FAIL --help does not show the '--json contract' block` |

Restored: all three ok. Full suite 8/8, including `test_pr_status_errored_review.sh`, which drives the same script.

Refutations attempted:

- *The `--help` change breaks a consumer that parsed a fixed range.* `test_verify_git_workflow_sections.sh` passes, and nothing else in the repo shells out to `pr-status.sh --help`.
- *`awk` on the leading comment block stops at the first blank line.* It stops at the first **non-comment** line; the header has no blank lines today. If someone introduces one above the contract block, `--help` truncates again — and the third assertion catches exactly that, which is why it is in the test rather than assumed.
- *The documented-block parser is brittle.* It keys on the `--json contract` heading, the three-space indent, and the sentence that ends the list. A future edit that adds keys *after* that sentence would not be seen — and the test then reports them as undocumented, i.e. it fails loudly rather than passing quietly. That is the correct direction to be wrong in.
- *This is documentation, so it needs no test.* It is a machine-readable interface with a silent failure mode. That is the case for a test, not against one.

One thing this does **not** fix: nothing stops a caller writing `.mergeStateStatus` and getting `null`. Only jq can see that, and it will not. The header now says so and points at `--watch` first, because in my case the entire hand-rolled loop was a substitute for a flag that already existed.

PR #183, opened 2026-08-13
docs(references): API-quota discipline, gh error-body capture trap, reset-rebuild verification

CybotTM:
Retro findings from the 2026-08-13 fleet-release-driver session (skill-repo-skill [#218](https://github.com/netresearch/skill-repo-skill/issues/218)/[#219](https://github.com/netresearch/skill-repo-skill/pull/219)), each observed live before being written down:

- **pull-request-workflow — the quota is one shared pool; act, don't re-preflight.** A heavy session killed the GraphQL secondary quota between "all gates verified" and the merge, taking `gh pr merge` and pr-status.sh with it while REST still answered. New subsection: one SHA-pinned REST merge on a verified unchanged head (`-f sha=$(git rev-parse HEAD)` — the 409 IS the freshness check, and it caught a hand-retyped SHA live), REST twins for branch delete, `gh api rate_limit` is exempt, stop known-answer watchers.
- **ci-cd-integration — `$(gh api … || echo "")` captures the ERROR BODY as data.** gh 2.97 prints HTTP error JSON to stdout without applying `--jq`; a fleet survey using the fallback-capture form classified never-released repos from their own 404 bodies. New subsection with the exit-code-gated capture pattern; applies to `glab api` too.
- **advanced-git — a reset-based rebuild commits the INDEX, not the worktree.** Editor writes after the last `git add` silently stay out of the collapsed commit; red CI was the first signal. New subsection: `git status --porcelain` + `git grep <artifact> HEAD` (bare, no pipe) before pushing a rewritten branch.

### DIFF
--- a/skills/git-workflow/references/advanced-git.md
+++ b/skills/git-workflow/references/advanced-git.md
@@ -2,6 +2,26 @@
 
 ## Rewriting History
 
+### After ANY reset-based rebuild: the commit takes the INDEX, not the worktree
+
+`git reset --soft <base>` + `git commit` is the standard way to collapse a
+branch into one clean commit — and it commits whatever is **staged**, which
+is the old tree, not the files as they lie on disk. Edits made after the last
+`git add` (every editor/tool write) silently stay out, the commit "succeeds",
+and only CI notices that the pushed tree is the stale one (2026-08-13: a
+history rewrite meant to purge two files shipped without the accompanying
+test/docs edits; the red CI on removed-file assertions was the first signal).
+
+After the rebuild, before pushing:
+
+```bash
+git status --porcelain          # anything listed = NOT in the commit you just made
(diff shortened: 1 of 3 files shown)


CybotTM (commented):
Review (fallback — Copilot failed on quota for the fourth time today; per convention the review is performed and documented by the operator): docs-only, three reference additions, each fact-checked against the 2026-08-13 session evidence. One claim corrected before merge: "GraphQL secondary quota" → REST and GraphQL are SEPARATE 5,000/h session-shared pools; the GraphQL pool died first while REST had headroom (`gh api rate_limit`: graphql 0, core 4,418). The error-body capture behavior was verified empirically against gh 2.97 (404 body on stdout, `--jq` not applied), the SHA-pin 409 caught a mistyped SHA live, and the reset-rebuild index trap was observed via red CI on the stale tree. markdownlint clean.

PR #184, opened 2026-08-14
fix(pr-status): say the Copilot quota wall once, clearly, without retries

CybotTM:
## Summary

The Copilot review quota is **monthly**: once the error body reads "reached their quota limit", no re-request on any PR succeeds until the monthly reset. `pr-status.sh` treated every failed review as a maybe-outage — advising one re-request per PR and letting `--watch` be re-armed — so the operator rediscovered the same wall on every pull request of the month (today: four PRs, each with its own re-request round and narration).

Changes:

- New `--json` field `copilot_quota_hit`, derived from the errored review body (`test("quota"; "i")`; the live error text is "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit").
- Decision ladder: on a quota error the advice is issued **once** and states plainly that the condition will not change this month, on this or any other PR; the retry command is withheld even on the first failure. Generic outage behaviour (one re-request, then self-review) is unchanged.
- `--watch`: the quota condition returns immediately with the same will-not-change-this-month notice instead of the generic failed-2x line.
- Tests: case 7c (quota on first failure → no cmd, MONTHLY advice present) and case 7d (a generic outage row must NOT trip the quota flag — the detector can stay quiet).

Verified against the live quota-failed PR [netresearch/typo3-docs-skill#76](https://github.com/netresearch/typo3-docs-skill/pull/76) in both one-shot and `--watch` mode; full `test_pr_status_errored_review.sh` and `test_pr_status_json_contract.sh` suites pass; `bash -n` and shellcheck clean.

## Test plan

- `bash tests/test_pr_status_errored_review.sh` — all pass (incl. new 7c/7d)
- `bash tests/test_pr_status_json_contract.sh` — all pass (new key documented)
- Live run against a quota-failed PR: NEXT carries the monthly notice, no cmd; `--watch` returns once with the unsatisfiable notice

### DIFF
--- a/skills/git-workflow/scripts/pr-status.sh
+++ b/skills/git-workflow/scripts/pr-status.sh
@@ -38,7 +38,7 @@
 #   base head headOid
 #   checks checks_settled threads unresolved_threads
 #   reviewDecision reviews_on_head has_review_on_head
-#   has_copilot_review_on_head copilot_review_errored copilot_error_count
+#   has_copilot_review_on_head copilot_review_errored copilot_error_count copilot_quota_hit
 #   requested_reviewers
 #   merge_methods auto_merge_allowed queue_active queue_entry
 #   rulesets rules_fetched required_contexts undispatched unsigned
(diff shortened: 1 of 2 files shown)


CybotTM (commented):
Reviewed the full diff (Copilot review unavailable — monthly quota exhausted, which is precisely what this PR teaches the script to say once instead of per PR). Verified: bash -n and shellcheck clean; test_pr_status_errored_review.sh all pass incl. new cases 7c (quota on first failure -> no retry cmd, MONTHLY advice) and 7d (generic outage does not trip the quota flag); json contract updated and its test passes; live-verified against quota-failed netresearch/typo3-docs-skill#76 in one-shot and --watch mode. Merging per maintainer instruction to land this fix (self-merge is documented practice here when no reviewer is available).

PR #189, opened 2026-08-15
fix(pr-status): carry the Copilot quota wall across PRs

CybotTM:
## The failure

`pr-status.sh` detects Copilot's review-quota wall from an errored review body on the PR being inspected — and only from there. The quota itself is per account and per calendar month, which the script's own comment already says ("once this is true, no re-request on any PR will succeed until the monthly reset"), but the evidence is per PR: a pull request nobody ever requested a Copilot review on carries no error row at all. On those PRs the NEXT ladder kept answering `request-review` together with a ready-to-run `gh api … requested_reviewers -X POST` command. Observed on 2026-08-14/15 in `netresearch/maint` #52 and #53: the operator ran the command the tool handed over, hours after the same account's quota had already been proven exhausted on another repo, and the request predictably never produced a review.

## The fix

The wall is remembered where the next invocation can find it. When a PR proves it (`copilot_quota_hit`), the run writes one marker file per calendar month to `${XDG_CACHE_HOME:-$HOME/.cache}/pr-status/copilot-quota-exhausted-YYYY-MM`, containing the PR that proved it and a timestamp so the verdict can be traced back. Every invocation reads the current month's marker before deciding.

The month is in the **filename**, never in the mtime — any later write would refresh an mtime and let a marker outlive the reset it describes — so a marker from an earlier month is ignored by name comparison alone and needs no cleanup. Writing is best effort throughout: an unwritable or full cache directory costs one wasted re-request, never a failed status query.

`copilot_quota_hit` keeps exactly its previous meaning, the evidence found on THIS pull request, so the existing per-PR detection still stands on its own. The new `copilot_quota_exhausted` is the verdict — evidence or marker — and it is what the NEXT ladder and the `--watch` exemption now read. Three branches change: the errored-review branch (unchanged behaviour when the evidence is local, marker-aware otherwise), the `copilot_code_review` ruleset branch, and the generic review gate — the last two being precisely the ones that used to hand back the POST command on a PR with no Copilot row of its own. No other NEXT outcome moves. The quota prose is hoisted into one binding used by all three, and it names which of the two sources the verdict came from plus the marker path, so an operator can tell a fresh reading from a remembered one and delete the file if it was recorded in error.

## Verification

`shellcheck` clean on both changed shell files; `pre-commit run --all-files` passes apart from a pre-existing `SC2148` on `.envrc`, untouched here. All eight `tests/*.sh` pass, including the `--json` contract test against the extended key list.

Six new cases in `tests/test_pr_status_errored_review.sh` cover the marker write, the generic outage that must NOT write one, both suppressed branches (with and without the ruleset), the previous-month marker being ignored, an unwritable cache directory, and the `--watch` exemption firing without any error row. Each was checked against a targeted mutation rather than assumed to bite: reverting the ladder to `copilot_quota_hit` fails Q3/Q4, dropping the month from the filename fails Q1/Q3/W5, removing the persist call fails Q1, and reverting only the watch-loop reads fails W5 alone.

The suite now pins `XDG_CACHE_HOME` into its own throwaway stub directory and clears the marker in `make_stub`/`make_stub_reviews`. The marker is the first piece of state that survives an invocation, so without that the suite would answer differently on a machine whose operator has hit the wall this month, would write into their real cache, and case W1 (watch must hold to timeout) would flip.

`references/pull-request-workflow.md` gains a paragraph on the quota case, which the surrounding text described only as a second-strike outage.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -138,6 +138,20 @@ yourself**: an outage may clear, but a quota ceiling does not clear by asking
 again, and the alternatives — merging unreviewed or waiting indefinitely on
 third-party infrastructure — are both worse than a self-review that says so.
 
+A body naming the **quota** is not an outage, so it costs no second strike:
+`pr-status.sh` withholds the retry command on the first failure and remembers
+the wall in
+`${XDG_CACHE_HOME:-~/.cache}/pr-status/copilot-quota-exhausted-YYYY-MM`. The
+quota is account-wide and monthly while the evidence is per PR — a pull request
+nobody ever requested a review on carries none at all — so without that file
+the next PR in the next repo is handed a `requested_reviewers` POST the same
+exhausted quota rejects (`netresearch/maint` #52 and #53, hours after the wall
+was proven elsewhere). Later runs read the marker, report
+`copilot_quota_exhausted: true` and answer with the self-review guidance and no
+command. The month is in the **filename**, so the marker stops applying at the
+reset rather than being aged out; delete the file to undo a verdict recorded in
+error.
+
 ## Check the Default Branch Before Operating
 
 Not every repo uses `main` — older repos often use `master`, and some use
--- a/skills/git-workflow/scripts/pr-status.sh
+++ b/skills/git-workflow/scripts/pr-status.sh
@@ -25,6 +25,17 @@
 # required checks concluded. Waiting for `pending == 0` means learning nothing
 # until the slowest matrix job ends, long after the first failure was visible.
 #
+# Copilot review quota
+#
+# The quota is per ACCOUNT and per MONTH, but the only evidence a single pull
+# request carries is an errored review body on its own head — a PR nobody ever
+# requested a review on carries none at all, and used to be answered with a
+# re-request command the exhausted quota rejects. So the wall proven on one PR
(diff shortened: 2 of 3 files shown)


CybotTM:
**Self-review** (Copilot review unavailable — monthly quota exhausted, which is precisely what this PR teaches the tool to remember; diff reviewed manually).

Checked on head f3dc001f: the design keeps `copilot_quota_hit` as per-PR evidence and introduces `copilot_quota_exhausted` as the verdict (evidence OR month-keyed marker), and every consumer that used to hand back the `requested_reviewers` POST now reads the verdict — the errored-review branch, the `copilot_code_review` ruleset branch, the generic review gate, and both `--watch` reads. Three details are exactly right: the marker is written from the evidence field, never the verdict, so a remembered wall cannot re-assert itself and the file always names the PR that actually proved it; the month lives in the filename rather than the mtime, so the marker expires at the reset by name comparison alone; and writing is best-effort throughout, so a read-only cache can never fail a status query. The rendered `why` distinguishes a fresh reading from a remembered one and names the marker path with the delete-to-undo instruction.

Verification is solid: shellcheck clean, all eight test files pass including the `--json` contract with the new key, six new cases each validated against a targeted mutation (ladder reverted, month dropped from the filename, persist call removed, watch reads reverted — each kills exactly the expected cases), and the suite now pins `XDG_CACHE_HOME` into a stub dir so it neither reads nor writes the operator's real cache. The reference doc gains the quota paragraph. All 22 checks green, 0 unresolved threads.

PR #193, opened 2026-08-15
docs: a worktree cleanup must check the directory's role, not only its branch

CybotTM:
Retro materialization (Learning-Id retro-20260815-cleanup-checks-role-not-only-state).

A cleanup sweep that classifies worktrees as removable by branch state alone (HEAD contained in `origin/main`, tree clean) also matches the repository's primary directory — whenever a `main/` worktree was switched to a feature branch and left there, and that branch has since merged. After the merge its state is indistinguishable from a disposable feature worktree. Removing it deletes the directory every other tool in the repository points at; the correct action there is a `git switch` back to the default branch.

Evidenced by a sweep over 171 bare-layout repositories holding 210 worktrees, of which 78 sat on a non-default branch: 12 passed the merged-and-clean test and 9 of those were primary directories.

The addition sits in `references/advanced-git.md` directly after the existing section on removing your own working directory, because both cover the same class: an operation that reports success and still hits the wrong thing. It comes with an eval covering the distinction.

### DIFF
--- a/skills/git-workflow/evals/evals.json
+++ b/skills/git-workflow/evals/evals.json
@@ -540,15 +540,15 @@
     "assertions": [
       {
         "type": "content",
-        "pattern": "(primary|main/|master/|default[- ]branch) (director|worktree)"
+        "pattern": "(?i)(primary|default[- ]branch|\\bmain\\b|\\bmaster\\b)[^\\n]{0,40}(director|worktree)"
       },
       {
         "type": "content",
-        "pattern": "git (-C [^ ]+ )?switch"
+        "pattern": "(?i)\\bswitch\\b[^\\n]{0,30}\\b(main|master|default)"
       },
       {
         "type": "content",
-        "pattern": "(not|never|do not|instead of) .{0,40}(remove|delete)"
+        "pattern": "(?i)(never|not|don't|do not|instead of|rather than)[^\\n]{0,50}(remov|delet)"
       }
     ]
   }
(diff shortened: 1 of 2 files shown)


CybotTM (commented):
Adversarial review across four dimensions, one refuter per finding defaulting to REFUTED: 26 findings over both PRs of this round, 18 confirmed, 8 refuted, none left unverified. Copilot remains unavailable (monthly quota, returns in September), hence this documented self-review.

The eval was inverted, and in both directions. Assertion 3 required the literal `remove` or `delete`; the reference text it was written for contains neither — it says "never-removable" and "not a removal". A correct answer therefore failed, while an answer removing every worktree including the primary one could satisfy the set. The assertions now bind the switch to the default branch, which is the step only a correct answer prescribes. Tested with a real regex engine against one correct and two wrong candidate answers: the old set failed the correct answer, the new set passes it and rejects both wrong ones. Also case-insensitive now, in line with the file's own `(?i)` convention.

The recipe omitted the fetch. Bare clones in this layout frequently lack `remote.origin.fetch`, so `origin/main` never moves and `merge --ff-only` silently does nothing while looking like it worked. The fetch is now in front of it, along with the case where another worktree already holds `main`.

The claim was also stated too broadly: git knows no "primary directory" role, the convention lives in the directory name of the bare-repository layout — which is exactly why a state-only classifier cannot see it. That is now said explicitly.

The numbers now carry a date and a population, and I measured them instead of writing them from memory: 171 bare-layout repositories with 210 worktrees, 78 on a non-default branch, 12 merged and clean, 9 of them primary directories. My first draft claimed "63 repositories", a number I had never measured.

One limitation that belongs in the PR text: this repository executes the evals nowhere — CI validates the JSON structure only, never a pattern. The new eval is therefore documentation of expected behaviour, not a gate.

PR #194, opened 2026-08-15
test: give the worktree-cleanup eval samples to check itself against

CybotTM:
Adopts the `samples` block from [skill-repo-skill#225](https://github.com/netresearch/skill-repo-skill/pull/225): one answer every assertion has to match, and four answers at least one assertion has to reject. Deterministic, no model call.

It closes the gap the review of [#193](https://github.com/netresearch/git-workflow-skill/pull/193) exposed: assertions were executed nowhere, CI validated the JSON structure only. Replayed against the original defect — the assertion demanding the literal `remove` while the reference text says "never-removable" — validation now names it: `assertion[2] does not match its own passing sample`.

Writing the samples exposed two further problems in the assertions themselves, both fixed here. One assertion required the negation to precede the removal verb, so an answer phrased "remove the merged ones, but not project/main" failed. Its replacement had the opposite flaw: it matched a bare stopword — `keep`, `leave`, `not` anywhere in the text — unbound from the decision the eval is about, so wrong answers passed it. Each of the three assertions now binds two things: the exemption to the primary directory, the remedy to the default branch, and the removal to the merged feature worktrees.

The failing samples were straw men that shared almost no vocabulary with a correct answer, so rejecting them proved only that the assertions were no longer inverted. Two more were added for the answers that were actually slipping through: one that switches correctly and then removes everything anyway, and one that removes the primary directory precisely because its branch merged.

Verified with `grep -qiE`, the matcher `run-ab-evals.sh` grades with, not with Python's `re`: the passing sample satisfies all three assertions, each of the four failing samples fails at least one, and all three patterns parse as POSIX ERE.

On ordering: this PR is green against today's validator (`samples` is an unknown key to it and is ignored) and is only really checked once skill-repo-skill#225 has merged — the `eval-validate` reusable is pinned to `@main`, so it picks the check up on the next run after that.

### DIFF
--- a/skills/git-workflow/evals/evals.json
+++ b/skills/git-workflow/evals/evals.json
@@ -540,22 +540,24 @@
     "assertions": [
       {
         "type": "content",
-        "pattern": "(?i)(primary|default[- ]branch|\\bmain\\b|\\bmaster\\b)[^\\n]{0,40}(director|worktree)"
+        "pattern": "(but not|except|other than|apart from|keep|leave|spare|exclude|never|do not|dont|not) [^.]{0,70}(main/|master/|primary|default.branch)|(main/|master/|primary|default.branch)[^.]{0,70}(is not|must not|never|stays|keep|leave|spare|exclude)"
       },
       {
         "type": "content",
-        "pattern": "(?i)\\bswitch\\b[^\\n]{0,30}\\b(main|master|default)"
+        "pattern": "(switch|check ?out) [^.]{0,40}(main|master|default)"
       },
       {
         "type": "content",
-        "pattern": "(?i)\\b(but not|except|other than|apart from|do not|don't|never|keep|leave|spare)\\b"
+        "pattern": "(remove|delete|prune) [^.]{0,70}(feature|merged|other)|(feature|merged|other) [^.]{0,70}(can go|be removed|removed|deleted)"
       }
     ],
     "samples": {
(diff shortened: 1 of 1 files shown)


CybotTM (commented):
Adversarial review across five dimensions, one refuter per finding defaulting to REFUTED: 29 findings over the three PRs of this round, 27 confirmed, 2 refuted, none left unverified. Copilot remains unavailable (monthly quota), hence this documented self-review.

The reviewer showed the assertion set still accepted realistic wrong answers, and that the samples proved less than they appeared to: the two failing answers shared almost no vocabulary with a correct one, so rejecting them only demonstrated the assertions were not inverted, never that they discriminate. The assertion I had rewritten in the previous round had swung to the opposite flaw — a bare stopword match on `keep`, `leave` or `not` anywhere in the text, unbound from the decision the eval is about.

Each of the three assertions now binds two things: the exemption to the primary directory, the remedy to the default branch, and the removal to the merged feature worktrees. Two further failing samples cover the answers that were slipping through — one that switches correctly and then removes everything anyway, and one that removes the primary directory precisely because its branch merged.

All of it verified with `grep -qiE`, the matcher the grader actually uses, not with Python's `re`: the passing sample satisfies all three assertions, each of the four failing samples fails at least one, and all three patterns parse as POSIX ERE — which matters, because `(?:...)` is not valid there and the first draft used it.

PR #195, opened 2026-08-15
feat: deny a forge body written in the chat language

CybotTM:
Repository artifacts are English. That rule existed in prose and did not hold: across one session, four PR bodies and four review comments went out in German — in `agent-engineering`, `git-workflow-skill` and `skill-repo-skill` — because the conversation was in German and the artifacts inherited the language without anyone deciding so. All eight had to be rewritten and PATCHed afterwards.

Adds a gate beside the hard-wrap one, sharing its body extraction (`_forge_bodies`, factored out of `forge_body_hard_wrapped`): a body is denied when German function words account for at least 8% of its prose across at least 6 distinct markers.

**What counts as prose.** Fenced blocks, inline code, indented blocks, blockquotes and quoted strings are removed before measuring, so an English body carrying a German error log, a UI label or a test fixture is judged on its own text. The first version of this branch denied exactly that body while its own message promised the opposite.

**What is covered.** `gh pr|issue|release create|edit|comment|review`, the `glab` equivalents, and `gh api` against a `/comments` or `/replies` endpoint with `-f body=` or `-F body=@file`. That last one matters: half the German text of the incident went out through `gh pr review` and through the review-reply endpoint, and the first version of this gate watched neither.

**Marker list.** `mit` is absent although it is a German word — it is also the licence every skill repository names, 18 times in one 63k-word English corpus. `das`, `von` and `hat` are gone for the same reason (DAS, von Neumann, HAT), as are `die`, `man`, `war`, `so` and `in`. A marker that fires on English is worse than a missing one in a gate that denies rather than nudges.

**Calibration**, on the artifacts that motivated it rather than on invented samples: the 8 German bodies from that session score 11 to 32 distinct markers at 12.7% to 23.6%; the 10 English ones, two of which quote German verbatim, all measure 0. The thresholds sit in the middle of that gap.

`FORGE_LANGUAGE_GATE_OFF=1` in front of the command exempts a repository whose own language is German — read off the command text, because the hook is its own process and would never see it as an environment variable, and matched as a leading assignment so that naming it inside a body cannot switch the gate off for that body. Commit messages are out of scope; this covers forge bodies only.


CybotTM (commented):
Adversarial review across five dimensions, one refuter per finding defaulting to REFUTED: 29 findings over the three PRs of this round, 27 confirmed, 2 refuted, none left unverified. Copilot remains unavailable (monthly quota), hence this documented self-review.

The gate's own denial message promised that quoting German inside an English body is fine, and the gate denied exactly that — a body carrying a German error log tripped it. Fenced blocks, inline code, indented blocks, blockquotes and quoted strings are now removed before measuring, and two tests cover the case with a realistic German stack trace and a PR about a German translation catalogue.

The coverage did not match the message either, and the gap contained the incident itself: half the German text went out through `gh pr review` and through the review-reply endpoint with `-f body=`, and the first version watched neither, nor any `glab` command. All three are covered now — which also means the gate finally watches the channel the PR-review rules mandate for addressing feedback.

`mit` left the marker list. It is a German word, and it is also the licence every skill repository names — 18 occurrences in one 63k-word English corpus. `das`, `von` and `hat` went for the same reason. In a gate that denies rather than nudges, a marker that fires on English is worse than a missing one.

Two robustness defects: a body file that is not valid UTF-8 raised out of the hook and took every other gate in the file with it, and the escape hatch was a substring test, so writing `FORGE_LANGUAGE_GATE_OFF=1` inside a body switched the gate off for that same body. It is now an assignment in front of the command, and a test asserts the difference.

Recalibrated on the same corpus after all of it: the 8 German bodies still deny at 11 to 32 distinct markers and 12.7% to 23.6%; the 10 English ones, two of which quote German verbatim, all measure 0.

One limitation worth stating: the gate lives in this plugin, so it takes effect after release and installation, not on merge. Until then the rule is carried by attention, which is what failed in the first place.

PR #196, opened 2026-08-15
docs: write down what building the language gate taught, and pin it

CybotTM:
Retro materialization (Learning-Id retro-20260815-deny-message-is-a-spec) from the round that shipped the forge-body language gate in [#195](https://github.com/netresearch/git-workflow-skill/pull/195). Several of that gate's defects were found by review rather than by me, and each generalises past that one gate.

`references/claude-code-hooks.md` gains three anti-pattern rows and a section on **the deny message as a specification**: whatever a denial promises is a required test case, and the exception a message carves out is the clause most likely to be wrong, because it is the one the implementation had to think about separately. The language gate is the worked example — of the three claims its first message made, the gate honoured one. A fourth clause entered the message only with the fix, which is the pattern rather than an exception to it: sharpening a promise is how the missing case gets named, so the tests are re-derived whenever the message changes. Also recorded: an exception read from the environment is inert in a hook, since the hook is its own process and never sees a `VAR=1 cmd` prefix; and a gate built from an incident must cover every command shape that incident used — that gate missed `gh pr review` and the review-reply endpoint, which carried half of it.

`references/pull-request-workflow.md`: the section on a stale body attributed the drift to elapsed time. The frequent trigger is a push answering a review, and it fires within the hour — the body still describes the approach the fix replaced, as happened on that PR the day it opened. Re-measure every number rather than copying it forward: the same PR claimed a "63k-word corpus" while the code comment beside the measurement said 60k. Both had been measured, of different corpora at different times, which is precisely why the disagreement is worth catching.

`references/ci-cd-integration.md`: ask the jobs API which step failed before reading any log — four log calls produced nothing about a failure whose step was called `Python lint`, and one API call named it. Then a section on why a green pre-commit run is not a green CI lint step, for two reasons. The run may not have been green at all: `files were modified by this hook` is a failure, and it is easy to discard when the output is filtered. And even a green run only proves what the *pinned* version thinks — here `.pre-commit-config.yaml` pins ruff v0.15.14 while CI pins `ruff@0.16.0`, and measured on the same file 0.15.14 reports `All checks passed!` where 0.16.0 fails on `SIM905`. Keep the two pins in parity; when they disagree the CI's version decides.

The marker-list calibration was prose in a code comment. It is now two tests that can fail: the nine words which also occur in English (`mit` is the licence every skill repository names) must stay out of `GERMAN_MARKERS`, and an English body about licences goes through the real gate rather than through a re-implementation of its thresholds. Verified by mutation — re-adding `mit` turns the suite red.

### DIFF
--- a/scripts/test_validate_git_command.py
+++ b/scripts/test_validate_git_command.py
@@ -18,7 +18,7 @@ from typing import ClassVar
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 
-from validate_git_command import german_prose
+from validate_git_command import GERMAN_MARKERS
 
 HOOK = Path(__file__).with_name("validate_git_command.py")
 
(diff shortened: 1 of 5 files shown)


CybotTM (commented):
Adversarial review across four dimensions, one refuter per finding defaulting to REFUTED: 29 findings over both PRs, 21 confirmed, 8 refuted, none left unverified. Copilot remains unavailable (monthly quota), hence this documented self-review.

The result is uncomfortable and worth stating plainly: nearly every confirmed finding was a false claim in prose I had written to teach the discipline of not making false claims. The dimension that recomputed the numbers and replayed the episodes against git history found that the CI section named the wrong cause outright. It said a pre-commit run was clean; the recorded run printed `files were modified by this hook`, which is a failure, and it had been read as success because the output was piped through the same `tail` the next sentence forbids. And "a repo's pre-commit set is not its CI lint set" is false for this repository — ruff is in `.pre-commit-config.yaml` at `rev: v0.15.14`, CI pins `ruff@0.16.0`, and measured on the offending file 0.15.14 says `All checks passed!` where 0.16.0 fails on `SIM905`. A version skew, not a missing hook, which also means the habit that section prescribed would not have caught it. Rewritten to say what happened and to prescribe keeping the pins in parity.

Two smaller corrections of the same kind: the worked example listed four clauses of a deny message, but the fourth was written into that message by the fix, so the table now shows the three that existed — which makes the point better, since sharpening a promise is how the missing case gets named. And the stale-body section asserted that one of two disagreeing figures was never measured; both were measured, of different corpora at different times.

The heaviest finding was against the test. It was supposed to pin the marker calibration and could not fail for that reason: it re-implemented the gate's thresholds instead of calling the gate, and an 8% share is structurally unreachable for multi-thousand-word reference files, so the corpus could never deny whatever the marker list contained. Its companion, meant to guard exactly that hole, could not see the thresholds either. Both are replaced by two that do fail — the nine English-colliding words must stay out of `GERMAN_MARKERS`, and an English licence body goes through the real gate. Verified by mutation: re-adding `mit` turns the suite red, which it did not before.

Also fixed: a rewrap of mine had touched 40 lines of neighbouring prose, which is churn this skill's own rules forbid; the file was restored and only my own paragraphs rewrapped to the surrounding 80 columns.

PR #198, opened 2026-08-18
docs(pr): reviewer-side pre-flight — read standing reviews before posting

CybotTM:
The pull-request-workflow reference covers the author and merger sides but has no reviewer-side section. This adds one: before composing review findings, fetch the PR's existing reviews and threads in the same batch as the diff — to dedupe against standing feedback, to recognize that a head pushed after a CHANGES_REQUESTED review may itself be the response to it (new findings can be regressions introduced by the requested changes), and to identify the standing blocking review as the actual merge gate. A subsection covers verifying subagent-reported file:line anchors against the fetched hunks before posting inline comments.

Origin: reviewing netresearch/usercentrics-widgets#143, a four-comment review posted straight from the diff duplicated a point of the standing CHANGES_REQUESTED review and missed that the head was the contributor's response to it; a subagent-reported line anchor (185) pointed at a context brace and had to be recomputed to 184.

Adds eval stub `review_pr_reads_standing_reviews_first` to evals.json.

### DIFF
--- a/skills/git-workflow/evals/evals.json
+++ b/skills/git-workflow/evals/evals.json
@@ -560,5 +560,19 @@
         "Remove the ones whose branches are merged. That includes project/main, since its feature branch merged as well."
       ]
     }
+  },
+  {
+    "name": "review_pr_reads_standing_reviews_first",
+    "prompt": "Review pull request #42 in this repository and post your findings as review comments",
+    "assertions": [
+      {
+        "type": "content",
+        "pattern": "(get_reviews|existing review|prior review|standing review|review threads?)"
+      },
+      {
+        "type": "content",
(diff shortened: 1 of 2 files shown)


CybotTM:
Copilot review is unavailable (monthly quota exhausted, account-wide), so per the documented fallback this got a manual review instead: docs-only change, two files, 50 insertions, no scripts or workflow surfaces touched. Verified: the new section lands between "If you catch yourself typing --squash" and "Review Thread Resolution" with balanced heading levels, markdownlint and check-json hooks passed at commit time, evals.json parses with the new `review_pr_reads_standing_reviews_first` entry matching the existing assertion schema, and the branch is based on current main (v1.26.0, 6d653a1).

PR #199, opened 2026-08-18
docs(pull-request-workflow): superseded-PR conflicts and silent Copilot-request drops

CybotTM:
## Summary

Two additions to `references/pull-request-workflow.md` from a five-PR review/merge sweep on 2026-08-18.

**A rebase conflict can mean the PR is superseded.** New section: when the main side of a conflict hunk already implements the PR's feature, the job changes from "resolve" to "re-scope to the delta main lacks, or close" — the reflexive keep-the-PR-side resolution silently reverts whatever main's version gained since. Concrete case: netresearch/jira-skill#194, where main's copy of the same stdin feature had a mention-gate call the PR predated; preferring the PR hunk would have merged green and dropped the gate.

**The Copilot review-request POST can be silently swallowed under quota.** Extends the existing quota section with the earliest tell: HTTP 200 with an empty `requested_reviewers` read-back, before any errored review exists. Five requests across four repos were spent this way before an errored review finally named the quota; the rule is to read back after the first request of a session and stop everywhere on an empty result.

## Came from

`/retro` sweep of the same session. Docs only, no script changes; markdownlint clean.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -152,6 +152,43 @@ command. The month is in the **filename**, so the marker stops applying at the
 reset rather than being aged out; delete the file to undo a verdict recorded in
 error.
 
+The quota wall has a second, earlier face: the `requested_reviewers` POST
+itself can be **silently dropped** — HTTP 200, but the response's
+`requested_reviewers` array is empty and a read-back seconds later still shows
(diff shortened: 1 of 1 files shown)


CybotTM:
Copilot review unavailable (monthly quota exhausted, account-wide). Self-reviewed the final diff against the patch-workflow checklist — commands verified runnable, incident facts checked against the session record — and merging under the /pr-finish mandate.

PR #200, opened 2026-08-18
fix(pr-status): let --watch hold through a consciously declined action

### LINKED ISSUE #165: pr-status.sh --watch returns instantly and repeatedly when the action is already set at invocation
## What happens

`--watch` exits on the first poll whenever `.next.action` is already one of the actionable values at invocation time. If that action cannot be cleared by retrying — a Copilot review that fails on a quota limit is the case I hit — `--watch` degrades into a no-op that returns the same line every time it is called, and cannot be used to wait for anything else.

`scripts/pr-status.sh` around the watch loop:

```bash
case "$act" in
  fix-ci|triage-ci|resolve-threads|request-review|rebase|resolve-conflicts|merge|blocked|none)
    echo "ACTIONABLE: $act"
    emit "$s"; exit 0 ;;
esac
```

Note the contrast with failing checks directly above it, which *are* de-duplicated across polls via `seen_fail`. Actions have no equivalent.

## Reproduction

On a PR whose head Copilot answered with a quota error (`Copilot was unable to review this pull request because the user who requested the review has reached their quota limit`), CI still running:

```
$ pr-status.sh -R netresearch/t3x-nr-llm 696 --watch
ACTIONABLE: request-review
  checks : 30 pass, 0 fail, 35 pending (17 running, 18 queued)

$ pr-status.sh -R netresearch/t3x-nr-llm 696 --watch      # immediately again
ACTIONABLE: request-review
  checks : 39 pass, 0 fail, 28 pending (19 running, 9 queued)
```

Both returned within a second. 28 checks were still pending and I had no way to wait for them through this tool, so I fell back to `gh pr checks <n> --watch` — which is exactly what the skill's own guidance steers away from.

The skill already documents that a repeated Copilot failure should be answered by reviewing it yourself rather than re-requesting (`references/pull-request-workflow.md`, "A failed Copilot review looks exactly like a delivered one"). So the operator can be in a legitimate, stable state where `request-review` is the correct `NEXT` *and* nothing about it will change — and `--watch` has nothing to offer.

## Why the current behaviour is defensible but incomplete

Returning on the first actionable event is the right default, and the README's argument against `until [ pending == 0 ]` loops is sound. The gap is that "actionable" is evaluated fresh on every invocation with no memory of what the caller already saw. A caller that has consciously decided not to act on `request-review` has no way to say so.

## Possible shapes (maintainers' call)

- De-duplicate actions the way `seen_fail` de-duplicates failures: capture `.next.action` at start and do not exit on that same value, only on a *change*.
- An explicit `--ignore-action <name>` (repeatable).
- A narrower `--watch-checks` that returns only on a check failing or the required contexts concluding.

## Environment

`git-workflow` 1.22.0 (installed via the netresearch marketplace), `gh` CLI, repository with an active `copilot_code_review` ruleset.

CybotTM:
Closes #165.

`--watch` exited on the first poll whenever `.next.action` was already one of the actionable values at invocation. When that action cannot be cleared by retrying — the reported case is a `request-review` pinned by an exhausted Copilot review quota — the watch degraded into a no-op that returned the same line on every re-arm and could not be used to wait for the still-pending checks.

Of the three shapes proposed in the issue, this implements the explicit `--ignore-action <name>` (repeatable). De-duplicating on the start value would change the default for everyone (a PR already at `merge` would hang until timeout instead of returning), and `--watch-checks` would duplicate `gh pr checks --watch`. The explicit flag keeps the first-event default and gives the caller who has consciously declined an action a way to say so.

Behavior with `--watch --ignore-action request-review`:

- never returns on the ignored action
- still returns on every other actionable event, and on new check failures unless the ignored action is the CI action itself (`fix-ci`/`triage-ci`)
- returns `SETTLED: NEXT is still the ignored action` (exit 0) once the checks settle with the ignored action still on top, instead of idling into the timeout
- rejects values outside the set of actions the loop returns on (a typo would otherwise behave as if the flag were absent), and refuses to run without `--watch`

The watch loop's action dispatch now derives from the same `ACTIONABLE` list the flag validates against; the previous duplicated `case` pattern is gone. The two UNSATISFIABLE advisories now point at the `--ignore-action` re-arm instead of `gh pr checks --watch`, which the skill's own guidance steers away from — falling back to it was part of the original report. `references/pull-request-workflow.md` and `references/merge-gate-watcher.md` are updated where they described the old dead end.

Tests: `tests/test_pr_status_watch_ignore.sh` covers the baseline return, the SETTLED exit, the hold-until-timeout on pending checks, and both flag validations. Verified as a guard by running it against the unfixed script: 6 of 13 assertions fail there, all 13 pass with the fix. The three existing pr-status/pr-merge test suites pass unchanged.

Second commit, addressing the independent review: the failing-checks suppression is narrowed to an ignored `fix-ci`/`triage-ci` (an ignored `resolve-conflicts` could previously swallow a fresh red check, since conflict/rebase outrank CI in the NEXT ladder), whitespace values for `--ignore-action` are rejected ("fix-ci triage-ci" slipped past validation as an `ACTIONABLE` substring), and the SETTLED message claims exactly what was checked. Four regression cases added; the conflict case fails on the first commit and passes on the second.


CybotTM:
Copilot review is unavailable this month (account review quota exhausted), so the diff was reviewed independently, per the fallback in `references/pull-request-workflow.md`. Initial verdict was **request changes**; all findings are addressed in d5435b6.

**Fixed — suppression broader than documented (important):** the failing-checks exit was suppressed whenever the current action was *any* ignored action, while the header, the inline comment, and the PR body all said "only when the ignored action IS the CI action". `resolve-conflicts`/`rebase` outrank `fix-ci` in the NEXT ladder, so an ignored `resolve-conflicts` swallowed a fresh red check and reported SETTLED (demonstrated with a controlled stub probe: CONFLICTING + FAILURE check). The guard now suppresses only for an ignored `fix-ci`/`triage-ci`.

**Fixed — test gap (important):** the flag's central positive promise ("still returns on every other actionable event") had no test. Three cases added: ignored `request-review` still returns on a red check; ignored `triage-ci` folds its own failures into SETTLED; ignored `resolve-conflicts` must **not** eat a failure — the last one fails on the first commit and passes on the second.

**Fixed — validation bypass (suggestion):** `--ignore-action "fix-ci triage-ci"` was accepted because the pair happens to be a substring of `ACTIONABLE`; whitespace in the value is now rejected, with a test.

**Fixed — SETTLED wording (suggestion):** message now claims exactly what was checked — `SETTLED: NEXT is still the ignored action` — since lower-ranked work can remain in the snapshot. Both reference files and the header follow.

**Declined — `--ignore-action` in the SKILL.md usage synopsis (suggestion):** the synopsis is consistently abridged (`--interval`/`--max-wait` are omitted too), SKILL.md sits at 499 of 500 words on the parallel #201 branch, and touching it here would conflict with that PR; the flag is fully documented in `--help` and both reference files.

Full battery after the fix: new suite 22/22 assertions, the three existing pr-status/pr-merge suites pass, `bash -n` and shellcheck clean.

CybotTM:
Re-review of d5435b6 by the same independent reviewer: **approve**, no new findings.

Re-verified: the CONFLICTING+FAILURE probe with `--ignore-action resolve-conflicts` now returns `ACTIONABLE: check failed` (previously SETTLED); the new test run against the first commit goes red on exactly the pinned regressions (6 assertions) and 24/24 pass on d5435b6; the old SETTLED wording is gone repo-wide; whitespace values die as documented. One noted caveat, no action needed: the "ignored request-review + red check" case also passes on the first commit (a red check flips NEXT to `triage-ci`, which even the broad guard let through) — it guards future breakage; the conflict case is the one that pins the original bug.

PR #202, opened 2026-08-18
feat: pr-status sees classic protection; --watch holds through the quota dead-end

CybotTM:
Two findings from the 2026-08-18 go-cron#399 session, each its own commit:

**Classic branch protection is invisible to pr-status.** The rules endpoint never shows classic protection's review settings, so go-cron#399 sat at BLOCKED with an APPROVED review on the head and every visible gate green — the approver was the last pusher and `require_last_push_approval` discounted the approval, but nothing named it. pr-status now lazily fetches the admin-only protection endpoint (only when BLOCKED or review-demanded, so the green path keeps two API calls), prints a `classic` line, exposes `classic_protection` in the JSON contract, and the NEXT ladder explains the discounted-approval case. Null for non-admin callers.

**--watch returned on the quota dead-end while checks were still pending**, forcing a manual `gh pr checks --watch` re-arm after every push (twice in one session). The watch now holds until required checks settle and only then returns the UNSATISFIABLE line; on settled CI the immediate return is unchanged.

Tests: the classic branch was exercised with a stubbed protection payload (fires on the deadlock state, inert on null); W3/W5 rewritten to assert the hold and verified failing against the previous script; new W6 guards the remaining immediate return. Full suite passes; JSON-contract test updated via the documented key list.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -28,7 +28,9 @@ bash "$GW/scripts/pr-status.sh" -R owner/repo 123 --json # for scripts and merge
 bash "$GW/scripts/pr-status.sh" -R owner/repo 123 --watch
 ```
 
-Two API calls, and the output ends in a computed `NEXT:` — rebase, fix-ci,
+Two API calls (a third, admin-only, when the PR is review-blocked — it reads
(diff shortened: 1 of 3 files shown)


CybotTM:
Copilot review remains quota-walled this month; shipped on self-review — behavior verified by the test suite in the diff (including the failing-on-old-script proof for W3/W5).

PR #204, opened 2026-08-18
docs(git-workflow): fetch a failed job's log while the run is still live

CybotTM:
The ci-cd-integration reference sends readers to the run's log archive for a failed job's real output — but that archive does not exist until the run completes (`gh run view --log` and the run-archive endpoint both answer "run … is still in progress"), which is exactly when you want the log of a matrix leg that already failed. The job endpoint works mid-run, with two traps now documented: `gh api` refuses a body containing terminal escape sequences unless `--allow-escape-sequences` is passed (the refusal is a single line with exit 0 — easy to read as an empty log), and the body arrives ANSI-colored, so the recipe pipes through an ANSI strip.

Also softens the section's implicit "job logs only carry runner preamble" into what it is — a per-job lottery: measured 2026-08-18 on netresearch/t3x-nr-vault (unit leg failed at 42% of a live run), the job-level body carried the full PHPUnit failure. Guidance: job endpoint first while the run lives, run archive once it is complete.

Source session: the t3x-nr-vault #303–#306 sweep, where discovering the flag cost several round trips.

### DIFF
--- a/skills/git-workflow/references/ci-cd-integration.md
+++ b/skills/git-workflow/references/ci-cd-integration.md
@@ -144,10 +144,11 @@ gh api "repos/$R/actions/jobs/$JOB/logs" --allow-escape-sequences \
 Without the flag the only output is
 `the response contains terminal escape sequences; pass --allow-escape-sequences
 to output it anyway` — one line, exit 0, easy to mistake for an empty log.
-Measured on 2026-08-18 (matrix leg failed at 42% of the run): the job-level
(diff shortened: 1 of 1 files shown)


CybotTM:
Review note: Copilot's review quota is exhausted for this month, so no bot review will arrive. Fallback per policy: docs-only diff, re-read critically before merge — one finding (an unverifiable '42% of the run' progress figure) was corrected in 34b1646 so the reference states only what was measured. No further findings.

PR #205, opened 2026-08-18
docs: a probe command is read-only; write the PR body in its own call

CybotTM:
Two rules from one session, both from failures that produced no error at the moment they happened.

**A probe command is read-only** (`references/advanced-git.md`). An exploratory one-liner comparing two revisions of a script carried `git stash -q`. It stashed the uncommitted work it was inspecting. Nothing failed — the probe printed its output and the tree was quietly a different tree, which is what makes this worth a rule rather than a note: a probe that reverts looks exactly like a probe that found nothing. Recovery was `git stash pop`, available only because the harness happened to print file-change notices. The section states the rule beside the existing selective-revert and selective-commit siblings, and gives the alternatives that make moving the tree unnecessary (`git show REV:path > /tmp/…`, a throwaway worktree).

**Write the PR body in its own call, after the push** (`references/pull-request-workflow.md`). A body citing a commit was composed before the push, so the hash was not yet resolvable on the remote and the call was refused. Because the push, the heredoc writing the body file, and `gh pr edit --body-file` were one shell invocation, the refusal discarded the heredoc — while the push had already run. The next command failed with `no such file or directory`, two steps downstream of the actual cause and reading like an unrelated problem.

The second half generalizes past PR bodies: a denied command loses every side effect it contained, and the parts that already ran do not roll back. So never bundle a file write with a push or an API call — keep state-changing steps separable, so a refusal costs the step rather than the scaffolding around it. The hash belongs in the body by lookup (`git rev-parse --short HEAD`), never typed from memory; a reader who follows a wrong hash finds nothing.

Docs only, no behaviour change.

### DIFF
--- a/skills/git-workflow/references/advanced-git.md
+++ b/skills/git-workflow/references/advanced-git.md
@@ -185,6 +185,36 @@ git merge --no-ff cherry-pick-fixes
 
 ## Stashing
 
+### A probe command is read-only
+
+Comparing two revisions, checking what a script used to do, reproducing a bug
+against an older build — that work is exploratory, and it must not move the
+working tree. Never put `git stash`, `git reset`, `git checkout --` or
+`git restore` inside a command whose purpose is to find something out.
+
+The failure is silent, which is what makes it worth a rule. A probe that
+stashes uncommitted work looks exactly like a probe that found nothing: the
+command prints its output, the tree is quietly a different tree, and the next
+edit lands on top of a state nobody chose. Recovery is `git stash pop`, but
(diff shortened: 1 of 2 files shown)


CybotTM:
Copilot review is unavailable — the account's monthly review quota is exhausted, so no bot review ran on this PR. Flagging that rather than letting the absence read as a clean review.

This PR came out of a `/retro` pass over the session that produced it, so the evidence for each claim is the transcript of the failure it describes, not a general principle. CI is green.

PR #206, opened 2026-08-18
feat(git-workflow): put the quota-fallback self-review on the record

### LINKED ISSUE #203: pr-merge.sh has no path to merge on the documented Copilot-quota fallback
When the monthly Copilot review quota is exhausted, pr-status.sh's own guidance (since f3dc001, the cross-PR quota marker) is: review the diff yourself, say in the PR that the bot review was unavailable, and decide on that. In a three-PR sweep today (netresearch/t3x-nr-vault#319–#321) that fallback was followed to the letter — an independent adversarial review per PR, findings fixed, a review-note comment on each PR — and `pr-merge.sh` still refused all three merges on the request-review NEXT, so every merge had to run as raw `gh pr merge --merge` outside the skill's gate. The tool that exists to be the safe merge path is unusable in exactly the situation its own status half documents a procedure for.

The source's rationale is recorded and understood: the tool cannot observe that a human read the diff, and an earlier "review-yourself" action was removed as unreachable-to-leave. An **explicit operator attestation** would respect that stance rather than contradict it — the tool still observes nothing, the operator asserts: e.g. `pr-merge.sh --attest-reviewed <evidence-url>` (accepted only while `copilot_quota_exhausted` is true and every other gate is green), where the evidence URL is the PR comment documenting the fallback review. The attestation could be echoed into the merge output so the assertion is on the record. Alternatively pr-merge could detect a review-note marker comment on the PR itself. Design call either way — hence an issue, not a PR.

Two adjacent observations from the same sweep, for the record: the quota-marker logic exists on `main` but not in the released 1.24.0 (the plugin cache), so sessions running the released version still get the pre-marker ladder incl. the pointless re-request command — a release would ship f3dc001/c9da33e/87d6126. And the marker file was present (`~/.cache/pr-status/copilot-quota-exhausted-2026-08`) while the 1.24.0 script ignored it, which is what made the two script generations look inconsistent within one session.

CybotTM:
Closes #203.

When the monthly Copilot quota wall stands, the documented fallback — review the diff yourself, note it in the PR, decide — ended outside the tooling: `pr-merge.sh` still refused the merge, so the operator merged with raw `gh pr merge`, past the very script that exists to be the safe path (three times in one sweep on 2026-08-18). Per the issue discussion, the agent can now tell the gate, and the gate can read the record back.

**pr-status.sh** fetches the PR's last 50 issue comments inside the existing GraphQL document (the two-API-call green path is unchanged) and reports `self_review_on_head` / `self_review_url` when the PR **author** posted a comment carrying the line `Self-review: <head-sha>` for the **current** head. The NEXT ladder honours the attestation only where it already calls the demanded review unsatisfiable — the quota wall, or two failed bot reviews on this head. With a live review path it changes nothing (the re-request command is still offered), a non-author comment never counts, and the next push invalidates it because the sha stops matching. A merge verdict resting on it says so and links the comment; the quota advice now names the exact marker line to post. This deliberately stays an explicit operator assertion the tool reads back, not a state it claims to observe — the same reasoning that removed the old "review-yourself" action (its regression, case 8, stays green).

**pr-merge.sh --self-reviewed** drives it end to end: refuses unless the demand is unsatisfiable AND the authenticated gh user is the PR author, posts the attestation comment when none exists yet, re-reads the gate, merges. Every other refusal stays exactly as strict; without the flag nothing changes.

Tests: seven new pr-status cases (attestation opens the gate under quota and under two-strikes; non-author, stale-sha, prose-only and no-wall cases all refused; the quota advice advertises the marker) and a new `test_pr_merge_self_reviewed.sh` with six cases (happy path incl. posted body, not-unsatisfiable refusal, non-author refusal, reuse of an existing attestation, unchanged no-flag behaviour, two-strikes acceptance). All new cases were run against the pre-change scripts and failed as expected; the full suite (10 sh + 1 py) is green, including the json contract test against the extended key list.

Docs: pr-status header (quota section + `--json` contract), pr-merge usage, a "Putting the self-review on the record" section in `references/pull-request-workflow.md`, and the SKILL.md usage line (word count still 499 ≤ 500).

## Review round

An independent adversarial review (see the review-note comment) confirmed the widening is contained and produced one Critical and three Important findings, all adopted in [da96d27](https://github.com/netresearch/git-workflow-skill/commit/da96d27): `--dry-run` no longer performs the gate-opening comment write; the flag keys on the refusing branch's `next.reason: bot-review-unsatisfiable` instead of account-global quota state (no false attestation onto a satisfiable classic-protection refusal); the quota advice names the marker via a placeholder so a pasted advice line can never mint an attestation; the accept side requires 12 sha chars (vanity-sha resistance); `CHANGES_REQUESTED`/`REVIEW_REQUIRED` keep the attestation inert; `comments(last:100)`; plus four more edge-case tests. Suites: 11 sh + 1 py, all green.

### DIFF
--- a/skills/git-workflow/SKILL.md
+++ b/skills/git-workflow/SKILL.md
@@ -76,7 +76,7 @@ Before merging: threads resolved, CI green (incl. annotations), rebased, signed,
 # Gate state, next action:
 ./scripts/pr-status.sh [-R owner/repo] [PR] [--json] [--watch]
 # Merge; refuses when the gate is shut:
-./scripts/pr-merge.sh [-R owner/repo] [PR] [--dry-run]
+./scripts/pr-merge.sh [-R owner/repo] [PR] [--dry-run|--self-reviewed]
 ```
 
 ---
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -174,6 +174,31 @@ quota). Read `requested_reviewers` back after the **first** request of a
 session; empty means stop requesting — everywhere — and go straight to the
 self-review path above.
 
+### Putting the self-review on the record (#203)
+
+"Review it yourself and say so in the PR" used to end outside the tooling: the
+review-note comment satisfied the policy while `pr-merge.sh` still refused the
+merge, so the merge ran as raw `gh pr merge` — three times in one sweep, past
+the very script that exists to be the safe path. The fallback is now a
+first-class input:
+
+```bash
+pr-merge.sh -R owner/repo 123 --self-reviewed
+```
+
+posts a PR comment whose body carries the line `Self-review: <head-sha>` (as
+the PR author — the flag refuses any other authenticated user), then re-reads
+the gate and merges. `pr-status.sh` reads the attestation back from the last
+50 comments and honours it **only** where the demanded review is one it
+itself calls unsatisfiable — the quota wall, or two failed bot reviews on the
+head. With a live review path the attestation changes nothing, a non-author
+comment never counts, and the next push invalidates it because the sha stops
+matching. This is an explicit operator assertion the tool reads back, not a
+state it claims to observe — the same reason the old "review-yourself"
+*action* was removed. The attestation comment is permanent PR history: post
+it only when the diff was actually reviewed, and say what was looked at in
+the review-note comment beside it.
+
 ## A Rebase Conflict Can Mean the PR Is Superseded
 
 When `NEXT: rebase` turns into a conflict, look at **what the main side of the
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -187,13 +187,19 @@ pr-merge.sh -R owner/repo 123 --self-reviewed
 ```
 
 posts a PR comment whose body carries the line `Self-review: <head-sha>` (as
-the PR author — the flag refuses any other authenticated user), then re-reads
-the gate and merges. `pr-status.sh` reads the attestation back from the last
-50 comments and honours it **only** where the demanded review is one it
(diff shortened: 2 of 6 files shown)


CybotTM:
**Review note:** Copilot's review quota is exhausted for this month, so no bot review will arrive. In its place an independent adversarial review agent (fresh context, read-only) reviewed the full diff with the explicit brief to find ways this change widens the merge gate too far. Verdict: the widening is well-contained (no path found where the attestation opens the gate while a live review path exists; every probed race fails closed), with one Critical and three Important findings — all adopted in 7c2990d:

- **Critical:** `--dry-run --self-reviewed` posted the real gate-opening attestation comment while only previewing the merge. The comment is now previewed, never posted, under `--dry-run` (new test).
- The flag keyed unsatisfiability on account-global quota state, so a satisfiable refusal (classic `require_last_push_approval`) could have received a factually false "unsatisfiable" attestation into permanent PR history. The refusing branch now stamps `next.reason: bot-review-unsatisfiable` and the flag keys on that alone (test pins the disagreeing-global-state trap).
- The quota advice embedded the exact accepting line; a terminal-wrapped paste could mint an attestation. The advice now uses a placeholder and never contains the accepting sequence (asserted).
- 8 hex chars is a 32-bit binding vanity-sha grinders defeat; the accept side now requires 12 (pr-merge posts the full oid).

Also adopted from the suggestions: `CHANGES_REQUESTED`/`REVIEW_REQUIRED` keep the attestation inert, `comments(last:100)`, and four more edge-case tests (marker mid-body, blockquoted paste, pending request outranks attestation, refusing re-read). Both suites green after the fixes; the reviewer had verified the GraphQL selection live against this very PR.

CybotTM:
Self-review: da96d278944c5cd42d16aefa098752246341eb44

PR #208, opened 2026-08-19
test(evals): migrate the 26 unchecked evals, drop the one this skill delegates

### LINKED ISSUE #207: evals/evals.json is never validated, and fails the current schema
This repo ships **two** `evals.json` and CI has only ever validated one of them.

```
skills/git-workflow/evals/evals.json   34 evals, last touched 2026-08-15   <- the one CI validates
evals/evals.json   31 evals, last touched 2026-08-10   <- never validated
```

The reusable `eval-validate.yml` picks the first match of `skills/*/evals/evals.json`, then `evals/evals.json`, and passes that single path to the validator. The second file has therefore never been checked.

Run against the current validator, the root file fails: its evals carry `id`, `name`, `prompt` and `expected_output`, but **no grading mechanism** — neither `assertions` nor `expectations`. That format predates the rule that every eval must be gradeable, so the file is inert: the A/B runner cannot score a single one of those 31 prompts.

**What this is not:** a regression. Nothing broke — the coverage was never there. I found it while fixing the same first-match-wins hole in the validator itself ([netresearch/skill-repo-skill#243](https://github.com/netresearch/skill-repo-skill/pull/243)), and measured six repos in the fleet that carry more than one eval file; three of them, this one included, have a root file in the old format.

**The decision is yours, and it is about content, not mechanics:**

- If those 31 prompts are worth keeping, they need `assertions` or `expectations` — and each assertion should be one the no-skill baseline actually fails, per the A6 rule.
- If they are superseded by skills/git-workflow/evals/evals.json, delete the file. The eval names in the two files are disjoint, so this is not a mechanical de-duplication either way.

I have deliberately **not** changed the reusable workflow to validate every file yet, because that would turn this repo red before the question above is answered. It is the natural next step once it is.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01CT41JfSGYEJJaBUZxg7xzu)_

CybotTM:
Closes #207.

The repo carried two `evals.json` and CI has only ever validated one: the reusable workflow takes the first match and passes that single path, so `evals/evals.json` has never been checked. It fails today — its evals carry `expected_output` prose and no grading mechanism, the format that predates the rule that every eval must be gradeable.

**Unlike the two sibling repos, this file is not a leftover.** Four of its 31 evals exist in the skill file under the same name. The other 27 do not exist there at all — and the newest of them are the most specific scenarios either file holds:

- an enqueued PR still reading `CLEAN`, so `mergeQueueEntry` is the thing to look at
- a required check sitting in `queued` with no runner, which is not `in_progress`
- `BLOCKED` with everything green → one unsigned commit anywhere on the branch
- a silent dequeue whose failure lives on `gh-readonly-queue/<base>/pr-<n>-<sha>`, invisible to every PR-scoped query
- `isRequired` being the only way to know whether a red check blocks
- ten PRs through one queue via a single integration branch
- `&& git commit` running after a resolver that exited 1, landing conflict markers

They were added to the **old** file after the migration, which is why the root file's last change (2026-08-10) postdates most of the skill file. Deleting it would have thrown them away.

So **26 move across** with assertions derived from what each `expected_output` recorded, and the stale file goes.

## Two judgements, stated rather than buried

- **`semantic_release_config` is not migrated.** This skill's own description says releases belong to `github-release`. An eval whose subject the skill explicitly delegates cannot discriminate *for this skill* — it would grade the model's general knowledge.
- **`conflict_resolution_reads_the_body` keeps the scenario, with a relaxed second assertion.** From the null-coalesce detail to "read the surrounding code per hunk": the skill teaches the general stance, not that specific consequence, and asserting more than the skill teaches makes an eval unpassable for reasons that are not the model's.

## How those two were found

Beyond `validate-evals.sh`, which checks ERE validity and structure but says nothing about discrimination: **every distinctive token the new assertions demand was grepped against `SKILL.md` and `references/`.** Three came back missing — two were the real mismatches above, the third a false alarm from my own matcher (`git.absorb` compared as a literal against "git-absorb" in the text).

`validate-evals.sh`: **125 passed, 0 failed**. The append kept the file's formatting — `git diff --numstat` is `372 0`.

Found while fixing the same first-match-wins hole in the validator itself ([netresearch/skill-repo-skill#243](https://github.com/netresearch/skill-repo-skill/pull/243)).

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01CT41JfSGYEJJaBUZxg7xzu)_


CybotTM:
Self-review — Copilot has no review quota left this month.

This is the one of the three where the sibling diagnosis did not transfer, so that is what I checked hardest. In `agent-rules` and `go-development` the root file was a leftover; here 27 of its 31 evals exist nowhere else, and the root file's last change (2026-08-10) is *newer* than most of the skill file — someone kept adding to the old location after the migration. Deleting it, which is what the other two PRs do, would have thrown away the sharpest scenarios either file holds: `mergeQueueEntry` on an enqueued PR that still reads `CLEAN`, `queued` versus `in_progress`, one unsigned commit shutting a green PR, the dequeue whose failure lives on `gh-readonly-queue/…`, `isRequired` as the only way to know whether a red check blocks.

**The check that produced the two corrections in this PR** is not `validate-evals.sh` — that one verifies ERE validity and structure and says nothing about whether an assertion can discriminate. I grepped every distinctive token the new assertions demand against `SKILL.md` and `references/`. Three came back missing:

- `semantic-release`: the skill's own description delegates releases to `github-release`. Migrating that eval would have graded the model's general knowledge under this skill's name, so it is dropped rather than carried.
- the null-coalesce consequence in the conflict eval: the skill teaches the general stance, not that specific outcome. Asserting more than the skill teaches makes an eval unpassable for reasons that are not the model's, so the assertion is relaxed to "read the surrounding code per hunk" and the scenario is kept.
- `git.absorb`: a false alarm from my own matcher comparing the pattern as a literal against "git-absorb" in the text. Worth naming, because two of three flags being real is exactly the ratio that makes a crude check worth running.

`validate-evals.sh`: 125 passed, 0 failed. `git diff --numstat` is `372 0` — the append did not reserialise the file.

22 checks pass, 0 unresolved threads, `mergeState=CLEAN`.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01CT41JfSGYEJJaBUZxg7xzu)_

PR #210, opened 2026-08-20
feat: upstream PR etiquette and a repo contribution preflight

CybotTM:
Two additions from contributing to a foreign repository, where the process is visible to people who did not ask for it.

## The etiquette section

`references/pull-request-workflow.md` gains **Upstream and Community PRs: the Maintainer Is Watching Every Push**.

In your own repository churn is private until someone looks. In someone else's, every force-push and body rewrite is the contribution as the maintainer experiences it.

The most expensive item is the first one: **a maintainer's question is a question**. On the branch this came from, a maintainer asked whether a second call site should also warn about something. That was read as an instruction, and it produced roughly 250 lines of diagnostics, a logger threaded into a collector, a traversal guard and a second directory walk — all deleted two hours later, because the correct answer turned out to be to remove the half-wired diagnostic rather than wire it everywhere. They had pointed at an inconsistency, not specified a feature.

The rest: land the minimal fix on its own and propose the improvement separately, so a bug fix does not wait on agreement about something else; answer a review in one comment rather than three; rewrite the body once, when the branch has its final shape; and do not restate in a PR what its linked issue already catalogues, since that pushes the reviewable part below the fold.

## The preflight script

`scripts/repo-contribution-preflight.sh` returns a repository's binding rules in one call. It reads and changes nothing. Sections are selectable with `--section docs|templates|packaging|ci|tools`.

On the repository this came from, the rules were: a 19-line `CONTRIBUTING.rst` that is a signpost to two other pages, `blank_issues_enabled: false` (so an issue must use a template), `/tests export-ignore` in every package (so published packages ship no tests), a sub-split workflow (so each package's `composer.json` is a real contract), a root `composer.json` with no production `autoload` at all (so any check run at the root scans nothing), and a `composer-require-checker` pinned at 3.5.1 that aborts on current `symfony/config` and prints empty output indistinguishable from a pass.

Each of those is one command. Together they are fifteen, which is why they get skipped — and then discovered with three artifacts already public and a maintainer watching the retrofit.

## Second commit: what the review changed

The first version of this PR — script and prose together — rested on the premise that a repository's rules *"are rarely in the README"*, and the script consequently never opened one. For small repositories the opposite is the common case: no `CONTRIBUTING` at all, and the whole contract in the README. Target branch, sign-off, the command CI runs. Such a repository was reported as having **no contribution docs**, which is the one answer a preflight must never give wrongly.

- **The README is read**, and its rule-bearing headings reported — **fence-aware**, because a heading regex is blind to code blocks and a README demonstrating `## Contributing` inside a fenced example would otherwise be reported as having that section. ATX and the underline form used by RST and setext are both handled.
- **`(none)` was a claim the script had not checked.** GitHub serves `CONTRIBUTING`, `CODE_OF_CONDUCT`, `SUPPORT`, `SECURITY` and both template kinds from the **owner's `.github` repository** when the repository itself has none, and those defaults bind identically. This script does not go to the network, so it now prints the query for that fallback instead of an absence it never verified.
- **Case-insensitive discovery** — `contributing.md` counts.
- **The full community-health set** — `SUPPORT`, `GOVERNANCE`, `SECURITY`, `CODE_OF_CONDUCT` — under the root, `.github/` and `docs/`.
- **The `PULL_REQUEST_TEMPLATE/` directory form**, plus the root and `docs/` locations for both template kinds.
- The link list **says how many it dropped** rather than truncating at eight silently.

The prose changed with it: the section no longer tells the reader to skip the README, and it now states the two things the script deliberately does not do — it never answers "there are no rules", and it does not treat a missing heading as a missing rule, since a three-line README can require a sign-off in running prose.

## Tests

New `tests/test_repo_contribution_preflight.sh` — 16 assertions over throwaway git repositories, no network, no fixtures on disk. **Every new assertion fails against the previous version of the script.** The three that still pass there are correct to: `CODE_OF_CONDUCT` was already in the old list, exit 0 always held, and the fenced-heading case passes vacuously because the old script read no README at all.

Two harness bugs surfaced while writing it, both of the kind that makes a test *lie* rather than fail, so they are named in the commit:

- A `newrepo` helper incrementing a counter handed every case the **same** directory, because it runs inside `$( )`. Absence assertions then pass without exercising anything.
- Piping the script into `grep -q` under `set -o pipefail` **inverts every assertion**: grep exits at the first match, the script dies of `SIGPIPE`, and the pipeline reports failure *because* the match succeeded. Output is captured into a variable and matched from there.

All 11 files in `tests/` pass. `verify-harness.sh --status`: Level 1 4/4, Level 2 3/3, Level 3 3/3. `shellcheck` clean on both files.

## One open question, unchanged from the first version

`SKILL.md` is at **499 of its 500-word cap**, so this script is still not in the `Verification` block where `verify-git-workflow.sh`, `pr-status.sh` and `pr-merge.sh` are listed. It is reachable only from the reference section. Listing it needs about fourteen words cut from somewhere else — a maintainer's call about what the entry point should carry, not one to make silently while fixing something adjacent.

_Assisted by claude-code:claude-fable-5 ([Session](https://claude.ai/code/session_0114KJz3vqq2WWfx4FUdmcss)) and claude-code:claude-opus-5 ([Session](https://claude.ai/code/session_01MNg1MysJVugv1xo2husknU))_

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -2740,14 +2740,13 @@ issue is filler that pushes the reviewable part below the fold.
 
 ### Read the repository before the first artifact
 
-Not the README — the parts that encode rules no prose mentions. See
-`scripts/repo-contribution-preflight.sh`, which returns them in one call:
-contribution docs and the pages they link, issue and PR templates, the CI
-matrix that actually runs, `.gitattributes` export rules, and the pinned tool
-versions the pipeline uses.
+The README **and** the parts that encode rules no prose mentions. Reading only the second half is a trap of its own: a small repository often has no `CONTRIBUTING` at all and states the entire contract in the README — target branch, sign-off, the one command CI runs, whether issues are wanted before PRs. A preflight that skips it reports "no contribution docs" about a repository that documented everything.
 
-Discovering `CONTRIBUTING.rst` after three artifacts are already public means
-retrofitting them in view of the people you are contributing to.
+`scripts/repo-contribution-preflight.sh` returns all of it in one call: contribution and community-health docs (`CONTRIBUTING`, `CODE_OF_CONDUCT`, `SUPPORT`, `GOVERNANCE`, `SECURITY`, found case-insensitively under the root, `.github/` and `docs/`) with the pages they link, the README headings that carry rules, issue and PR templates including the `PULL_REQUEST_TEMPLATE/` directory form, the CI matrix that actually runs, `.gitattributes` export rules, and the pinned tool versions the pipeline uses.
+
+Two things it deliberately does not do. It never answers "there are no rules": when a repository has no contribution docs of its own, GitHub serves the owner's `.github` repository defaults instead, and those bind identically — so the script prints the query for that fallback rather than an absence it did not check. And it does not treat a missing heading as a missing rule: a three-line README can require a sign-off in running prose.
+
+Discovering `CONTRIBUTING.rst` after three artifacts are already public means retrofitting them in view of the people you are contributing to.
 
 ## Full PR Lifecycle Checklist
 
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: 9477b9ffd1b310562184805cbe9d4a0c5938e73a

PR #215, opened 2026-08-21
fix(pr-status): an approval on the head satisfies the bot-review demand

### LINKED ISSUE #214: pr-merge.sh --self-reviewed writes a false "unsatisfiable review" attestation onto already-approved PRs
`pr-merge.sh --self-reviewed` posts the "the review is unsatisfiable" attestation onto pull requests that **already carry a valid approval on the current head**. The comment is permanent and its central claim is false.

## Evidence

`netresearch/t3x-nr-llm#860`, merged 2026-08-21. Head at merge: `83b35361a145`.

```
11:26:07Z  github-actions[bot]  APPROVED   commit=83b35361a145   ← the merge head
11:31:55Z  pr-merge.sh          posted "Self-review: 83b35361a145"
```

The approval landed on the current head **five minutes before** the script decided the review demand was unsatisfiable and wrote:

> The review this pull request demands is unsatisfiable (Copilot quota wall or repeated bot failures on this head).

It was satisfiable. It was satisfied.

**Seven for seven.** Every PR I merged that day (`#858`, `#860`, `#862`, `#863`, `#865`, `#869`, `#871`) carries at least one such attestation, and every one of them shows `reviewDecision: APPROVED` at merge time. `#863` carries two and `#865` three — one per head after a force-push.

## Why this is more than noise

`pr-merge.sh` is careful about exactly this hazard, and says so:

```
# Keyed on the reason the REFUSING BRANCH stamped, never re-derived from the
# account-global quota state: … with the monthly quota marker set on this
# machine a global test would post a factually false "unsatisfiable"
# attestation into permanent PR history there.
```

That guard works — the reason really is read from the refusing branch. The gap is one level up: `pr-status.sh` returns `next.action = request-review` with `reason = bot-review-unsatisfiable` **even when `reviewDecision` is already `APPROVED` on the current head**, so `--self-reviewed` proceeds down a branch that should never have been reachable.

The consequence is a permanent, on-the-record claim that a review could not be obtained, on a pull request where one demonstrably was. Anyone auditing why a change merged without a human review reads a statement that is not true.

## What I think the fix is, stated as a proposal rather than a diagnosis

Before `request-review` is emitted at all, `reviewDecision == "APPROVED"` on the current head should short-circuit to `merge`. If GitHub's own gate reports `mergeStateStatus: CLEAN` **and** `reviewDecision: APPROVED`, there is no review to demand — the host has already answered the question the action exists to ask.

I have not read `pr-status.sh` closely enough to say whether the approval is consulted and then overridden, or never consulted on that path. That distinction matters for the fix and I did not want to guess at it.

## Not this, and worth separating

There is a *second*, unrelated way this surfaces, and conflating them cost me an hour: when a stacked PR's base merges, GitHub retargets the child and **dismisses its approval** (`The base branch was changed.`). Then `REVIEW_REQUIRED` is correct and the tool's refusal is correct — only its *stated reason* names the quota rather than the dismissal. That case is covered by #213; this issue is the other one, where no dismissal happened and the approval is live.

CybotTM:
Closes #214.

## What was wrong

The three copilot branches in `pr-status.sh` asked whether a **Copilot** review had landed on the head, and never whether anything else had. So a PR that `github-actions` had already `APPROVED` on the current head still came out as `request-review` with `reason: bot-review-unsatisfiable` — which is precisely the input `pr-merge.sh --self-reviewed` keys on to post its permanent *"the review this pull request demands is unsatisfiable"* attestation.

Seven PRs merged in one sweep on `netresearch/t3x-nr-llm` carry that comment while showing `reviewDecision: APPROVED` at merge time. On [#860](https://github.com/netresearch/t3x-nr-llm/pull/860) the approval landed on the merge head at 11:26:07Z and the attestation was posted at 11:31:55Z — five minutes after the review the comment says could not be obtained.

## The fix, and why it is the branch's own position

The bot branches now additionally require that **no approval sits on the current head**. That is what the ruleset branch already says it is enforcing, in its own `why` text: the rule *"does not block the merge, since a Copilot review does not count toward required approvals; the demand here is the never-merge-unreviewed policy, not a host gate"*. An approval on the head satisfies that policy. GitHub reports `CLEAN` in this state — a gate that genuinely blocks arrives as `BLOCKED` and is answered further up the ladder, ahead of these branches.

Gated on the **approval list**, not on `has_review_on_head`. The latter is true for any non-author review including a `COMMENTED` one, and a failed Copilot review *is* an ordinary `COMMENTED` row — testing it would let the error satisfy the policy it exists to enforce. This is not an argument I am asking you to take on trust: substituting `($s.has_review_on_head|not)` for the new predicate makes **case 15** fail (`staleness warning dropped by a COMMENTED reply`), which is how I checked it.

## Tests

Cases 13 and 16 asserted `request-review` for exactly this state. An earlier pass corrected the *wording* those branches emit and left the *action* — that is the half #214 is about, so both now assert `merge`, each with the reason recorded at the case.

New case 20 covers the leg the reported PRs actually took: the wall known only from the month marker, **no Copilot row on the head at all**, an approval present. New case 20b is its twin — same marker, same missing row, approval on an *older* commit — so the dead end is still reported where it is real, and case 20 cannot be satisfied by deleting the marker leg.

Both new cases fail without the script change:

```
case 20: quota MARKER + no copilot row + APPROVED on head — no false dead end
  FAIL next.action: expected 'merge', got 'request-review'
  FAIL next.reason absent: expected 'null', got 'bot-review-unsatisfiable'
```

Full suite with the fix: `118 ok, 0 FAIL` in `test_pr_status_errored_review.sh`; the four sibling `test_pr_*` files unchanged at `3/32/24/24 ok`. `verify-harness.sh --status`: Level 1 4/4, Level 2 3/3, Level 3 3/3. Pre-commit passed (shellcheck, markdownlint, whitespace, EOF, merge-conflict).

## Documentation

`references/pull-request-workflow.md` claimed a satisfiable refusal *"can never receive a false unsatisfiable attestation"*. The keying on the stamped reason is necessary and was not sufficient, since the refusing branch itself stamped it here — the passage now says so and names the added condition.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -195,7 +195,13 @@ honours it **only** where the refusing branch stamped
 `next.reason: bot-review-unsatisfiable` — the quota wall, or two failed bot
 reviews on the head; the flag keys on that reason, never on the account-global
 quota state, so a satisfiable refusal (say, classic `require_last_push_approval`)
-can never receive a false "unsatisfiable" attestation. With a live review path
+can never receive a false "unsatisfiable" attestation. That keying is necessary
+and was not sufficient: until #214 the refusing branch itself stamped the reason
+while an `APPROVED` review sat on the very same head, and seven approved PRs in
+one sweep got the attestation anyway. The bot branches now also require that no
+approval is on the current head — checked against the approval list rather than
+`has_review_on_head`, which a failed Copilot review satisfies by being an
+ordinary `COMMENTED` row. With a live review path
 the attestation changes nothing, a non-author comment never counts, a human
 `CHANGES_REQUESTED` or a host-required approval keeps it inert, `--dry-run`
 previews the comment without posting it, and the next push invalidates the
--- a/skills/git-workflow/scripts/pr-status.sh
+++ b/skills/git-workflow/scripts/pr-status.sh
@@ -648,7 +648,22 @@ evaluate() {
          # forever. The tool cannot observe "a human read the diff", so it no
          # longer pretends to: it keeps reporting the honest state and only
          # stops handing over a retry command a quota ceiling will reject.
+         # $no_current_approval gates this branch and the two below it for the
+         # same reason the generic gate at the bottom tests has_review_on_head:
+         # the demand is the never-merge-unreviewed POLICY, as the ruleset
+         # branch says in its own why-text, and an APPROVED review on this head
+         # satisfies it. Without the guard a PR that github-actions has already
+         # approved still reports request-review, and pr-merge.sh
+         # --self-reviewed then writes "the review this pull request demands is
+         # unsatisfiable" into permanent PR history — a claim that is false
+         # while the approval sits on the same SHA (#214).
+         #
+         # Not has_review_on_head: that is true for any non-author review
+         # including a COMMENTED one, and the Copilot error rows ARE COMMENTED,
+         # so testing it would let a failed bot review satisfy the policy. The
+         # approval list is the one predicate an error row cannot answer.
          elif ($needs_copilot and $s.copilot_review_errored
+               and $no_current_approval
                and ($s.has_copilot_review_on_head | not)
                and ($self_attested | not)
                and (($s.requested_reviewers|map(test("copilot";"i"))|any) | not)) then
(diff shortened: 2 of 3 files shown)


CybotTM:
Self-review: 22f09f500cabdd3005f6349d8cda0e555ff86558

Issue #216, opened 2026-08-21
SKILL.md does not list repo-contribution-preflight.sh — needs fourteen words, or a decision not to


CybotTM:
`skills/git-workflow/scripts/repo-contribution-preflight.sh` is not listed in `SKILL.md`, while the three other scripts in that directory are. It is reachable only from `references/pull-request-workflow.md`, which a reader gets to after already deciding what to do — the entry point is where a tool is chosen, and this one is invisible there.

## Why it was left out, and why that is not the same as a decision

`SKILL.md` sits at **499 of its 500-word cap** (`wc -w` on `main`), enforced by the `Validate skill repo structure` pre-commit hook. Listing the script the way the others are listed costs roughly fourteen words:

```
# What this repository expects, before the first artifact:
./scripts/repo-contribution-preflight.sh [--repo <dir>] [--section <name>]
```

So it needs fourteen words cut from somewhere else in `SKILL.md`. That is a judgement about what the entry point should carry, not a mechanical edit — which is why #210 shipped without it and asked instead. The question went unanswered through two rounds of work on that PR, and #210 has now merged, so the ask would otherwise be buried in the body of a closed pull request.

## What closing this needs

A decision on one of three, not a patch written first:

1. **Cut fourteen words** somewhere in `SKILL.md` and list the script beside the others.
2. **Raise the cap** if 500 is a convention rather than an external requirement — worth checking which it is before assuming it cannot move.
3. **Accept the current placement** and say so here, so the next person reading the `Verification` block does not re-open it.

Option 3 is a legitimate outcome. The reference section genuinely does describe when to reach for the script, and an entry point that lists everything stops being an entry point.

## Not to be confused with

This is only about discoverability from `SKILL.md`. The script itself, its README and community-health coverage, and its tests were settled in #210 — including the fence-aware README scan and the removal of the unchecked `(none)` claim.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01MNg1MysJVugv1xo2husknU)_

Issue #209, opened 2026-08-19
Two different script sets both declare 1.26.0 — 23 unreleased commits on main, including the self-review attestation


CybotTM:
Two installations of this skill on one machine ship different scripts while declaring the same version, so `pr-merge.sh --self-reviewed` exists in one and not in the other. A consumer has no way to tell which copy is running.

**Measured** (one machine, 2026-08-19):

| Path | pr-merge.sh | `--self-reviewed` | declared version |
|---|---|---|---|
| `~/.claude/plugins/cache/netresearch-claude-code-marketplace/git-workflow/1.26.0/skills/git-workflow/` | 193 lines | absent | `1.26.0` (SKILL.md + plugin.json) |
| `~/.agents/skills/git-workflow/` (and `~/.claude/skills/git-workflow` → symlink to it) | 254 lines | present | `1.26.0` |

`pr-status.sh` differs too (different md5); `spec-cleanup-guard.sh` is identical in both.

**Cause** — the cache is not mispackaged. `git show v1.26.0:skills/git-workflow/scripts/pr-merge.sh` is the 193-line version without the flag, so the cache matches its tag exactly. `main` carries 254 lines with the flag and still declares `1.26.0` in both `SKILL.md` and `.claude-plugin/plugin.json`. There are **23 commits on main since the v1.26.0 tag** (2026-08-17 → 2026-08-19), among them `feat/self-review-attestation` (#206) and `feat: put the quota-fallback self-review on the record` (#203). Features landed without a version bump, so one number now names two different artifacts.

**Impact, measured rather than assumed.** With the account-wide Copilot review quota exhausted, `pr-merge.sh` refuses every PR with `NEXT: request-review`. The documented way past that wall is `--self-reviewed` — which the tagged copy does not have. In one session that meant roughly a dozen PRs merged by hand with `gh pr merge` instead of the supported attestation flow, across netresearch/t3x-nr-llm, t3x-nr-llm-compat, typo3-demo, nr-landingpage, t3x-nr-mcp-agent, t3x-nr-repurpose and t3x-cowriter. The failure mode is silent: a missing flag reads exactly like a missing feature, and the invocation that ran the older copy was a version-pinned cache path that looked correct.

A second session on the same machine reproduced the same split and initially read it as a packaging error in the cache — the tag comparison above is what separates the two explanations.

[...]

PR #218, opened 2026-08-21
feat(scripts): every script answers --version and names its own path

CybotTM:
Refs #209 — the half of it that a release cannot close.

## What #209 measured, and what is still open

Two installations of this skill on one machine declared **the same version** while shipping different scripts, so `--self-reviewed` existed in one and not the other. Both numbers were truthful; `main` had simply moved without a bump.

The release fixed that instance. It did not make the next one visible, and #209 named the gap itself: *"a consumer cannot currently answer 'which copy am I running'."* The cost was measured, not assumed — a rejected flag reads exactly like a feature that was never built, and roughly a dozen PRs were merged by hand instead of through the attestation flow.

## What this adds

All seven scripts answer `--version`:

```
$ pr-merge.sh --version
pr-merge.sh 1.27.0
path: /home/you/.agents/skills/git-workflow/scripts/pr-merge.sh
```

The **path is not decoration**. The number alone is exactly what failed in #209, where both copies said `1.26.0` and meant different things. The path is what tells two installations apart.

The version is read from the `SKILL.md` **next to the script** — never from a checkout elsewhere, never from a git tag. A cached copy has to report the number it was packaged with, or the answer is worse than no answer.

It works where nothing else does: no repository, no network, no `gh` on `PATH`. And with no `SKILL.md` beside it, it reports `unknown` and exits 0 rather than failing — refusing there would make the flag useless in precisely the broken installation it exists to identify.

## Why seven copies of a function instead of a shared library

A sourced helper has one failure mode this feature cannot afford: if the library is missing from an installation, the version query itself breaks — at the exact moment it is needed. There is no `source` anywhere in this repository today, so a self-contained function also matches what is already there. Seven copies of a short function is the cheaper trade.

## Tests

`tests/test_scripts_report_version.sh`, 25 assertions.

**Case 3 is the property the flag exists for**: a second copy in another directory, with its own `SKILL.md` declaring `0.0.1-other`, must report *its own* number and *its own* path. An implementation that read the version from the repository root, or from `git describe`, would pass cases 1 and 2 and reproduce #209 exactly.

The suite reads the frontmatter with `python3` rather than reusing the `awk` the scripts use — a test that reuses the implementation cannot catch the implementation being wrong.

Case 4 pins the missing-`SKILL.md` behaviour, case 5 the stripped-`PATH`-outside-a-repository case, case 6 agreement with `.claude-plugin/plugin.json`.

**Every assertion fails against the previous scripts — 0 of 25 pass:**

```
FAIL merge-gate.sh reports the version: expected 'merge-gate.sh 1.27.0', got ''
FAIL pr-merge.sh reports the version: got 'pr-merge: unknown flag: --version'
FAIL pr-status.sh reports the version: got 'pr-status: unknown flag: --version'
```

All 12 files in `tests/` pass. `verify-harness.sh --status`: Level 1 4/4, Level 2 3/3, Level 3 3/3. `shellcheck` clean on all seven scripts and the new test.

## Not in this PR

`SKILL.md` is untouched and still at its 500-word cap, so the scripts remain unlisted there. That is #216 and needs a decision about what the entry point should carry, not a drive-by edit while fixing something adjacent. `--version` is documented in `references/pull-request-workflow.md` instead.

This also does not bump any version. There are now unreleased commits on `main` again, including the merge-gate fix from #215 — which is #209's other half, and a maintainer's call about when to cut a release.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -211,6 +211,20 @@ state it claims to observe — the same reason the old "review-yourself"
 it only when the diff was actually reviewed, and say what was looked at in
 the review-note comment beside it.
 
+### Before believing a script cannot do something: ask which copy is running
+
+Every script in `scripts/` answers `--version`, and prints the resolved path beneath it:
+
+```
+$ pr-merge.sh --version
+pr-merge.sh 1.27.0
+path: /home/you/.agents/skills/git-workflow/scripts/pr-merge.sh
+```
+
+The version is read from the `SKILL.md` next to the script, never from a checkout elsewhere, so a cached copy reports the number it was packaged with. The path is there because the number alone is not enough: two installations on one machine declared the same version while shipping different scripts, and `--self-reviewed` existed in only one of them (#209). The flag was reported as missing — which reads exactly like a feature that was never built — and about a dozen PRs were merged by hand instead of through the attestation flow.
+
+So when a documented flag is rejected as unknown, that is a question about the installation before it is a question about the tool. `--version` needs no repository, no network and no `gh`: it is the one thing that must still answer when everything else is broken.
+
 ## A Rebase Conflict Can Mean the PR Is Superseded
 
 When `NEXT: rebase` turns into a conflict, look at **what the main side of the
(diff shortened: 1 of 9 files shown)


CybotTM:
Self-review: 209376d4bae18ba6ef25429ea74938c2f2d9d3df

PR #220, opened 2026-08-21
fix(scripts): --help stops at the header instead of a hand-kept line number

CybotTM:
Refs #218 — what reviewing that change turned up, landing separately because #218 merged first.

## Two scripts documented themselves with a line number, and both had drifted

`--help` in `repo-contribution-preflight.sh` and `signing-preflight.sh` printed the file header with a fixed line range. Neither range still matched its header:

| Script | Range | Header ends | Result |
|---|---|---|---|
| `repo-contribution-preflight.sh` | `2,28` | line 25 | printed **three lines too many**, including `set -uo pipefail` as documentation |
| `signing-preflight.sh` | `2,30` | line 31 | **dropped the last line** of its own usage |

The first one is mine: I widened that range in #210 when the header grew, and overshot. The second is the more interesting failure, because nobody notices it — documentation that is missing looks exactly like documentation that was never written.

Both now use the self-terminating form the other scripts already use: print comment lines until the first line that is not one. A hand-kept range drifts the moment anything is inserted above the code, which is precisely what #218's `skill_version()` did to it.

## The test asserts the property, not a number

Case 7: **no `--help` output contains source code**. Asserting the line count instead would just be a second number to keep in step with the first.

It fails against the previous script:

```
FAIL repo-contribution-preflight.sh --help leaks no source: expected '0', got '1'
```

Scoped to the four scripts that *handle* `--help`. The other two treat it as a positional argument, and asserting there would pin unrelated behaviour rather than this one.

## One more thing the review checked

#218 put the `--version` guard at the top of every script, several of which run under `set -e`, and shipped tests for the **new** path only. Nothing covered the path that was already there.

Checked now: all seven behave exactly as they do on `main` when invoked without `--version`, and the two guard-style scripts return the same exit codes as before — `merge-gate.sh` 0, `verify-git-workflow.sh` 1. An AND-OR list is exempt from `set -e`, so the guard is sound; it simply had no evidence behind it.

## Verification

All 12 files in `tests/` pass, including against the freshly released **1.28.0** — `test_scripts_report_version.sh` reads the number from `SKILL.md` at run time rather than pinning it, so the release bump moved through it without an edit. `shellcheck` clean at the pre-commit hook's severity (the hook runs stricter than `-S warning`; the one `SC2016` is suppressed at the line with a comment saying why — `$1` there is regex matching the literal string `case "$1"`). `verify-harness.sh --status`: Level 1 4/4, Level 2 3/3, Level 3 3/3.

### DIFF
--- a/skills/git-workflow/scripts/repo-contribution-preflight.sh
+++ b/skills/git-workflow/scripts/repo-contribution-preflight.sh
@@ -65,7 +65,7 @@ while [ $# -gt 0 ]; do
         --repo) REPO="${2:?--repo needs a directory}"; shift 2 ;;
         --section) SECTION="${2:?--section needs a name}"; shift 2 ;;
         --version) skill_version; exit 0 ;;
-        -h|--help) sed -n '2,28p' "$0"; exit 0 ;;
+        -h|--help) awk 'NR>1 && /^#/ {sub(/^# ?/,""); print; next} NR>1 {exit}' "$0"; exit 0 ;;
         *) echo "unknown argument: $1" >&2; exit 2 ;;
     esac
 done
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: a1dfff8a42cfd60d3aeeae168ae2aeb323b042c6

PR #221, opened 2026-08-21
docs(skill): list the preflight script, and say where a fact must live to be seen

### LINKED ISSUE #216: SKILL.md does not list repo-contribution-preflight.sh — needs fourteen words, or a decision not to

CybotTM:
Closes #216.

## The literal ask

`repo-contribution-preflight.sh` is now listed in `SKILL.md` beside the other three scripts, together with `--version`. The file is at **486 words**, from 499.

The room came from moving **detail** into references that already carry it — the full conventional-commit type list into `commit-conventions.md`, the per-framework install commands into `git-hooks-setup.md` — and from dropping a `## When to Use` section that restated the frontmatter `description` almost verbatim. That description is read *before* the body loads, so repeating it in the body spends the tighter budget on an already-answered question. No capability was removed; the routing to `github-release` and `github-project` stays.

## The question behind it

> *the most important thing is, the agent skips the skill because a feature is missing from the description/frontmatter or from SKILL.md body … dunno if we need to stick to 1-level-references only!?*

`AGENTS.md` now states which surface a fact has to be on for the model to see it at all:

| Surface | When it enters context | What a gap there costs |
|---|---|---|
| `description` | always, in the skill listing | **the only true skip** — the skill is never invoked, so nothing further is read |
| `SKILL.md` body | on invocation | a blind spot: the skill runs, the model does not know the thing exists |
| `references/*.md` | only when the model chooses to read one | nothing, until it is needed |

Two numbers that matter for the first row: an entry's combined description text is **truncated at 1,536 characters**, and the listing as a whole is capped at about **1% of the context window** — on overflow, descriptions are shortened or dropped, starting with the least-used skills. Hence *key use case first*. This description is at 641 characters, so it has room.

**Scripts belong in the body.** They are executed, never loaded, so one line costs almost nothing and is the only chance the model has of knowing one exists — which is exactly what #216 was about.

## One level of references

The documented shape is flat: `SKILL.md` is overview and navigation, supporting files sit beside it, and each is referenced from `SKILL.md` *"so Claude knows what each file contains and when to load it"*. The `Reference Files` table's **Content Triggers** column is that mechanism, not decoration.

So no second hop. `SKILL.md` → `references.md` → `topic.md` puts the third file behind a door with no sign on it: nothing says what it holds or when it matters, so the model has to open the middle file speculatively and then guess again. **Split by topic at the first level** — `topic-a.md` and `topic-b.md` both in the table, each with its own trigger.

## Two findings recorded, not fixed here

Neither belongs in this diff, and both are worth knowing before someone else walks into them.

**The enforced budget is not the documented one.** The gate counts **500 words over the whole file**, frontmatter included. The published guidance is *"Keep `SKILL.md` under 500 lines. Move detailed reference material to separate files."* Words are a much tighter and differently shaped budget than lines, and charging the frontmatter to it puts the description in competition with the instructions — the inversion that produced this issue. This body is 76 lines against an allowance of 500.

**`Build/Scripts/validate-skill.sh` is a copy that nothing invokes.** The gate that actually runs is the `validate-skill` pre-commit hook from `netresearch/skill-repo-skill` (pinned at `v1.22.0`). I changed the local copy first and the commit was still rejected by the real one — editing it changes no gate, and the two are free to drift.

Both are being raised where they belong rather than patched around here.

## Verification

`validate-skill` (the real hook) passes: 0 errors. All 13 files in `tests/` pass. Pre-commit clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,6 +41,66 @@ baked-in defaults it *does* flag the dogfooded design spec, by design).
 - **Intermediate / working** — must never reach the base branch:
   `docs/superpowers/`, `claudedocs/`, `docs/working/`, ad-hoc `*.plan.md`.
 
+## Where a fact has to live for the model to see it
+
+Three surfaces, three different moments. Putting something on the wrong one is
+why a capability that exists still does not get used.
+
+| Surface | When it enters context | Budget |
+|---|---|---|
+| `description` in the frontmatter | always, in the skill listing | combined description text truncated at **1536 characters**; the listing itself is capped at ~1% of the context window |
+| `SKILL.md` body | when the skill is invoked | published guidance is **under 500 lines**; the gate here enforces **500 words over the whole file** (see below) |
+| `references/*.md` | when the model chooses to read one | none, they cost nothing until read |
+
+What follows from that:
+
+- **A missing capability in the `description` is the only true skip.** The model
+  never invokes the skill, so nothing further is consulted. Put the key use case
+  first: when many skills are installed the listing overflows its budget and
+  descriptions get shortened or dropped, starting with the least-used skills.
+- **A missing capability in the body is a blind spot, not a skip.** The skill
+  runs, and the model does not know the thing exists.
+- **Scripts belong in the body.** They are executed, never loaded, so listing one
+  costs a line and buys the only chance the model has of knowing it is there.
+
+### The enforced budget is not the documented one
+
+The `validate-skill` pre-commit hook comes from `netresearch/skill-repo-skill`,
+and it counts **500 words over the whole file**, frontmatter included. The
+published guidance is *"Keep `SKILL.md` under 500 lines"*. Words are a much
+tighter budget than lines, and charging the frontmatter to it means the
+`description` -- the one surface that decides whether the skill is used at all
+-- competes with the instructions for the same allowance.
+
+The practical consequence is visible in this repository's history: it sat at
+499 of 500 and left a script out of `SKILL.md` rather than spend fourteen words
+on it. Note also that `Build/Scripts/validate-skill.sh` here is a COPY that
+nothing runs -- editing it changes no gate.
+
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: d838e58537752184da6760ebcb00d4fded8e1cc9

PR #223, opened 2026-08-21
docs(git-workflow): check-runs lists one entry per run, not per check

CybotTM:
`ci-cd-integration.md` recommends `commits/<sha>/check-runs` for finding run ids, which is what it is good at. This adds what it is **not** good at.

A second workflow run on the same commit — a reopened pull request, a `workflow_dispatch`, a schedule firing — adds another set of check-runs beside the first, and the old ones keep their old conclusion:

```
security / SAST (Opengrep): success   22:12:34
security / SAST (Opengrep): success   22:12:31   <- previous run, same SHA
```

Measured on a pull request whose fuzz preflight was red until an upstream reusable workflow was fixed: reopening produced a green run, the endpoint kept listing the red one, and a monitor built on it reported "2 failing" while nothing was failing.

## The distinction that matters

`gh run rerun` does **not** do this — it updates the existing check-run in place. Verified on two pull requests where a rerun turned a red check green: one entry each, conclusion `success`. So the rule is *new run vs. rerun*, not "the API duplicates entries", which is how I first heard it and would have written it down wrong.

The added jq keeps only the newest run per name; checked against the affected pull request, it reports 0 failing where the naive count reports 2.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

### DIFF
--- a/skills/git-workflow/references/ci-cd-integration.md
+++ b/skills/git-workflow/references/ci-cd-integration.md
@@ -75,13 +75,24 @@ check-runs beside the first. The old ones keep their old conclusion, so a
 counter over that endpoint reports failures that no longer exist:
 
 ```
-security / SAST (Opengrep): success   22:12:34
-security / SAST (Opengrep): success   22:12:31   <- previous run, same SHA
+All security checks:           success   22:13:53
+All security checks:           failure   22:10:01   <- previous run, same SHA
+fuzz / Preflight (event gate): success   22:11:32
+fuzz / Preflight (event gate): failure   22:09:34   <- previous run, same SHA
 ```
 
 Measured on a pull request whose fuzz preflight was red until an upstream
-workflow was fixed: the reopen produced a green run, and the endpoint kept
-listing the red one, so "2 failing" was reported while nothing was failing.
+workflow was fixed, then closed and reopened. On that one commit:
+
+| Query | Failures reported |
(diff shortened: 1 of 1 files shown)


CybotTM:
The Copilot review quota is exhausted for this month, account-wide, so no bot review is available on this head. I read the diff myself instead, and that pass found a defect worth recording here rather than quietly amending.

The original example listed two `security / SAST (Opengrep)` entries, both `success`. It demonstrated that the endpoint duplicates entries, but not the consequence the section is about — a stale `failure` outliving the run that produced it. Re-measured on the same commit, the names that actually carry different conclusions across runs are `All security checks` and `fuzz / Preflight (event gate)`. Counting every check-run reports 2 failures there; newest-run-per-name and `gh pr checks` both report 0.

The correction is 5840c12f. It also adds the part that makes the trap hard to spot: checks that were green in both runs duplicate as well, so a spot check on a green name shows two identical entries and looks harmless.

Not merging this — that decision is not mine to make.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

CybotTM:
Self-review: 5840c12fb91a98469a3eba22c0210323698a65de

PR #224, opened 2026-08-22
docs(watcher): name-miss, empty SHA and a stale ref all read as "not there"

CybotTM:
Three ways a post-merge check reports "not there" when the thing is there, and the state is really an artefact of how it was looked for. All three came out of one session's watchers on 2026-08-21.

## A check run is named after the job

Searching `commits/<sha>/check-runs` for a *workflow's* name finds nothing, and the natural conclusion — the run does not exist for this commit — is wrong.

`netresearch/typo3-demo` run `32536686040` is `Deploy (Update)` (`event=workflow_run`, `check_suite_id=88200944968`); its check run `96938832284` is named `deploy`, same check suite, and its `html_url` points back at that run's job. Reproduced 4/4 there and 2/2 on `netresearch/ofelia`, where `Verify Release` appears as `Verify / Verify Release`.

## `head_sha=` filters server-side — and vanishes when empty

`actions/runs` accepts the SHA as a query parameter; the file previously only warned against client-side filtering of a windowed list. The parameter has one sharp edge worth stating: **an empty variable drops the filter silently** and the API returns the unfiltered list — 4506 runs measured that way on one repository, against 0 for a syntactically valid unknown SHA and 0 for outright garbage. A watcher whose SHA lookup failed then reports a stranger's run as the watched commit's, and because the extraction is not empty, a value guard cannot catch it. Hence the `[ -n "$SHA" ]` line before the loop.

The window defect itself is already documented here for `gh run list`. What is new is its behaviour **inside a loop**: it works on the first tick, because the commit is still recent, and goes blind later — and quiet is indistinguishable from "still running".

## A file read right after a merge can return pre-merge content

A watcher fired "the trigger is NOT on main" the moment a PR merged; a read at the merge commit a minute later showed it present at the expected lines. Address the merge commit, not the branch ref.

## `ready_for_review`, and the conclusion that does *not* follow

`pull-request-workflow.md` gains the type-set trap: `ready_for_review` is not in the `pull_request` defaults, so a workflow carrying a `draft == false`-gated auto-approval never re-runs when the draft is lifted, and the PR sits at `REVIEW_REQUIRED` with nothing red.

The tempting follow-up — "so never open the fix for this as a draft" — is **false**, and the section says so. For `pull_request`, GitHub reads the workflow file from the PR's merge ref, so a PR that adds the trigger already has it in force. Verified on [netresearch/ofelia#442](https://github.com/netresearch/ofelia/pull/442) and [netresearch/git-workflow-skill#156](https://github.com/netresearch/git-workflow-skill/pull/156), both of which introduced their own workflow and ran it at their own head SHA. The trap belongs to `pull_request_target`, which loads from the base branch.

## What an adversarial pass removed before it shipped

An earlier draft of this change asserted that a `workflow_run`-triggered deploy has **no** entry under `commits/<sha>/check-runs`, and used that to justify needing the runs API. Measurement refuted it — the entry is there under the job name. The claim had been reasoned, not measured; it is now replaced by the first section above, which is the more useful finding.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_0144iD1P22LotW8rxmxrNGro)_

### DIFF
--- a/skills/git-workflow/references/merge-gate-watcher.md
+++ b/skills/git-workflow/references/merge-gate-watcher.md
@@ -214,6 +214,60 @@ gh api repos/$R/commits/$SHA/status      --jq '{state, total:(.statuses|length)}
 
 **Pre-existing red ≠ your regression.** If a post-merge gate (e.g. SonarCloud "Quality Gate failed" on N Security Hotspots) is red, check the *prior* base commit before owning it: `gh api repos/$R/commits/<prev-sha>/check-runs --jq '.check_runs[]?|select(.name=="<gate>")|.conclusion'`. Identical red on the parent + a diff that touched no relevant code = a pre-existing backlog to report, not a regression to fix.
 
+### A check run is named after the JOB, not the workflow
+
+Looking for a workflow by its own name in `commits/$SHA/check-runs` finds nothing, and the natural conclusion — "that run does not exist for this commit" — is wrong. The entry is there under the **job** name.
+
+Measured on `netresearch/typo3-demo`: workflow run 32536686040 is `Deploy (Update)` (`event=workflow_run`, `check_suite_id=88200944968`); its check run 96938832284 is named `deploy`. Same check suite, and the check run's `html_url` points back at `/actions/runs/32536686040/job/96938832284`. Reproduced 4/4 there and 2/2 on `netresearch/ofelia`, where `Verify Release` appears as `Verify / Verify Release`.
+
+So match on the check suite or on the job names you expect, not on the workflow name — and never infer absence from a name miss.
+
+### When you want run-level state, filter server-side with `head_sha=`
+
+The workflow run's own `status`/`conclusion` (as opposed to its jobs') lives on the runs endpoint. Do **not** reach back for a windowed list and filter it yourself — `actions/runs` takes the SHA as a query parameter and filters on the server:
+
+```bash
+[ -n "$SHA" ] || { echo "no SHA — refusing to watch"; exit 1; }
+gh api "repos/$R/actions/runs?head_sha=$SHA&per_page=50" \
+  --jq '.workflow_runs[] | "\(.name)=\(.status)/\(.conclusion // "-")"'
+```
+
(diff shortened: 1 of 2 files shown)


CybotTM:
The bot review was unavailable: Copilot is out of quota, account-wide and monthly. The one entry on this PR sits on an older commit and is an error body, not a review; there is no approval on `1f9ee654`. What follows is what I checked.

**The central claim of the first section is one that replaced a wrong one of mine, so I re-measured it myself rather than trusting the report that corrected me:**

```
$ gh api repos/netresearch/typo3-demo/actions/runs/32536686040
  name=Deploy (Update)  event=workflow_run  check_suite_id=88200944968

$ … /commits/59a4e5cb…/check-runs | jq 'select(.check_suite.id==88200944968)'
  name=deploy  id=96938832284
  url=…/actions/runs/32536686040/job/96938832284
```

The earlier draft of this change asserted the opposite — that a `workflow_run` deploy has *no* check-run entry — and used it to justify the `head_sha` section. That was reasoned, not measured. It is gone.

**One claim I softened after the review pass, not because of it.** The stale-read paragraph said `contents/…?ref=<branch>` is "served from a cache that lags the merge by seconds". I did not test that mechanism; I only observed that the branch ref answered stale and the merge commit answered correctly. The text now says exactly that and marks the mechanism as unestablished — in a section whose whole subject is not mistaking how something looked for what happened.

**Claims carried over from the verification pass and not personally re-run:** the 4506-vs-0 measurement behind the empty-`$SHA` warning, and the two PRs cited for the `pull_request` merge-ref behaviour ([ofelia#442](https://github.com/netresearch/ofelia/pull/442), [#156](https://github.com/netresearch/git-workflow-skill/pull/156)). Both are checkable from the text.

**Repo gates:** `validate-skill.sh` (the version the pre-commit hook pins) reports 0 errors; the warning count is unchanged from the pre-patch baseline, so this introduces none. Fence parity checked. No `SKILL.md` change — the additions are sections inside an already-listed reference.

What I cannot claim: that the sections read well to someone who has not hit these failures. A second pair of eyes is still worth having when the quota returns.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_0144iD1P22LotW8rxmxrNGro)_

CybotTM:
Self-review: 1f9ee654e0d5cea8e0715907d0b291af232ee861

PR #227, opened 2026-08-25
docs(pull-request-workflow): return a PR to draft when work resumes

CybotTM:
## Problem

The PR Lifecycle section already covers opening every PR with `--draft` and the mechanics of the Draft → Ready edge, but it treats draft as a birth condition. There is no rule for the case that actually recurs: work resuming on a PR that already reads "ready for review" — a rebase onto a moved base, a round of review fixes, another commit of any kind. `gh pr ready --undo` appears only as an unexplained line in the command list, which reads as a remedy for a PR opened non-draft by mistake rather than as a step in the normal loop.

## Why it matters

"Ready for review" is a standing request for a maintainer's time against a specific head, and mid-work heads do not deserve it. Between the first fix commit and the last, such a PR advertises for review a state you already know is incomplete — sometimes one you know is broken, when the work is a response to a reviewer's finding. Reviewers who look during that window spend attention on a diff that is about to change, and a green CI run on an intermediate head reads as an endorsement of work that is not finished.

## Change

Adds the return edge to the state diagram, the rule with its reason, the tell that it was skipped (recent commits labelled as review fixes while `draft: false` held throughout), and the `gh`/`glab` commands. The command-list entry now says when it applies.

## Observed

phpDocumentor/guides#1344, 2026-08-24. The PR sat at "ready for review" through a rebase and two rounds of fixes answering a reviewer's finding, one of which the finding had shown to be broken. It took two corrections before the state was set right, and the rule that should have covered it only spoke about `gh pr create`.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -1125,12 +1125,12 @@ Draft → Ready for Review → Changes Requested → Approved → Merged
 
 **Draft is a state the PR returns to, not one it only starts in.** Opening every PR with `--draft` is the well-known half. The half that gets missed: when work resumes on a PR that is already "ready for review" — a rebase, a round of review fixes, another commit of any kind — convert it back *before the first push*, with `gh pr ready --undo <n>` (`glab mr update <iid> --draft`). Mark it ready again as a separate step, once the checks are green and the user has asked for it.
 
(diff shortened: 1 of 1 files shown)


CybotTM:
Copilot answered the ruleset's review request on `601524f2` with "unable to review … reached their quota limit" and no comments, so no bot review exists for this head. Reviewed it myself instead.

The review was not a formality — it caught the check command shipped with the new rule. `gh pr view --json` spells the field `isDraft`; `draft` belongs to the REST payload and `gh api`. Copied as written it returned `Unknown JSON field: "draft"`. Fixed in 601524f, which also names the discrepancy so the next reader does not trip on it from the other side, and drops a sentence that argued for the practice rather than describing it.

Verified on the head: `pre-commit run` over the changed file passes, `scripts/verify-harness.sh` reports Level 3 COMPLETE with 0 errors, `scripts/spec-cleanup-guard.sh` reports clean, and the corrected command was run against this PR before being committed. Branch is not behind `main`. No review threads.

Self-review: 601524f265b1367ff964e5a2835b3414b1f15d67

PR #229, opened 2026-08-25
docs(pull-request-workflow): two preconditions a stack has to meet

CybotTM:
## Problem

The stacked-PR material covers what happens once a stack exists: merge order, retargeting, `--delete-branch` closing the child, approvals lost when the base merges. Two things that decide whether the stack can exist at all were not written down, and both were paid for in one afternoon.

**A base must be a branch in the target repository.** Contributing from a fork, the feature branch lives in the fork, so upstream has nothing to point at:

```
$ gh pr create --base feature/x --head myorg:follow-up
Base ref must be a branch (createPullRequest)
```

The check is one call — `gh api "repos/$UPSTREAM" --jq '.permissions.push'` — and it belongs before the branch is built, not after the PR is refused.

**A split is only real if both halves stand on their own.** A test and the CI change that lets it pass read like two things and are one. Worse, the coupling is usually already proven at the moment the split is proposed: if you diagnosed a CI failure where one half causes it and the other fixes it, you have your answer and the split is a promise you cannot keep.

## Change

Two sections ahead of the existing stacked-PR material, since both are preconditions for it:

- *A stack needs push access to the repository the PRs target* — the constraint, the error it produces, the one-line check, and what the options actually are on a fork (single PR, or a follow-up that waits and targets the default branch). Also names the trap of proposing a split and a stack in the same breath, where only the merge order is deliverable.
- *Before splitting a PR, run each half without the other* — build each half, run the gates on it, and check your own diagnosis first.

## Observed

phpDocumentor/guides#1345, 2026-08-25. A review round grew the PR past its subject, so the extra work was split onto its own branch and committed — then the upstream PR could not be opened (`push: false` on the base repo), and the halves turned out to be inseparable anyway: any test loading the class hits the CI failure the other half fixes, in every suite. Branch, PR and description were rebuilt around the split and then rebuilt again to undo it.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -1759,6 +1759,49 @@ The second way PR-green diverges from main-green needs no `push:`-gated job: a P
 - Rebasing before merge does **not** close the window — the config PR can land after your rebase. Only a merge queue or a "require branches to be up to date" rule makes it structurally impossible; where neither is on, expect this occasionally on active repos.
 - Consequence: watching the post-merge run on `main` is part of the merge, because this failure class is visible nowhere else. When it fires, diff `main`'s history for config-touching merges since your branch's last CI run and fix forward against the new config — it is mechanical, not an investigation into which check was wrong.
 
+### A stack needs push access to the repository the PRs target
+
+A pull request's base must be a branch **in the repository the PR is opened
+against**. Stacking PR2 on PR1's branch therefore requires that branch to exist
+there — which it does when you push branches to the target repo, and does not
+when you contribute from a fork. Upstream has no `feature/x`; it lives in your
+fork, and `gh pr create --base feature/x` answers:
+
+```
+Base ref must be a branch (createPullRequest)
+```
+
+Check before you plan around a stack, not after you have built it:
+
+```bash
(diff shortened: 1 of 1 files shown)


CybotTM:
Copilot answered the ruleset's request on `e7528d50` with "unable to review … reached their quota limit" and no comments, so there is no bot review on this head. Reviewed it myself.

It caught one thing. The note quoted `Base ref must be a branch` as though that were GitHub's answer; the answer has four clauses, and a reader matching their own output against one of them will not recognise it. One of the others actively misleads — `No commits between …` is false in the sense it suggests, the branches do differ, and checking that is exactly the detour the note exists to prevent. Fixed in e7528d5: all four quoted, with the operative one named.

Verified on the head: `pre-commit run` over the changed file passes, `scripts/verify-harness.sh` reports Level 3 COMPLETE with 0 errors, `scripts/spec-cleanup-guard.sh` reports clean. Branch is not behind `main`. No review threads. Reference text only — no `SKILL.md`, plugin manifest or version surface touched.

The `.permissions.push` check the section recommends was run against a real repository today (`phpDocumentor/guides`, `push: false`), which is what produced the case this documents.

Self-review: e7528d500f563d5b0017acf3eb75f30b21c29a03

PR #230, opened 2026-08-25
docs: running a tool from a stale worktree, and two exit-code traps

CybotTM:
Two findings from a `/retro`, both about failures that present as something other than what they are.

## `advanced-git.md` — running a tool from a stale worktree

The "A stale worktree is a stale source" section covered *reading* a stale tree. It did not cover *running* from one.

A script merged an hour ago does not exist in a reference worktree that has not been fast-forwarded, so invoking it by path fails with `exit 127` — which reads as a broken command, a bad `PATH`, a typo, anything but "the tree is old". It happened twice in one session, the second time on a script that had been merged forty minutes earlier by the same session that then failed to run it.

The remedy belongs at the end of a merge rather than at the start of the next debugging session, and it is two lines. `--ff-only` is deliberate: a reference worktree that cannot fast-forward has local commits and has stopped being a reference worktree, which is worth finding out loudly. The note also says why the moment is easy to miss — after a merge the *branch* worktree is usually deleted, which is exactly when scripts start being invoked from `main/`.

## `code-quality-tools.md` — capturing an exit code under `set -e`

Two traps in shell test runners, placed next to the existing command-substitution gotcha because they share its shape. Both fail in the same misleading direction: **every case collapses at once**, so the runner looks wholly broken rather than subtly wrong.

- `env VAR=1 my_func` execs a *program* and never sees the shell function — 16 passing cases became 16 failures at `127`.
- `set -e` inside `$( )` aborts the subshell before the `echo $?` that was going to capture the code, so a runner whose cases are *supposed* to fail prints its header and nothing else.

Neither is caught by shellcheck. SC2163 does fire inside the corrected block — on `export "$assignment"`, which is the one line there that is correct as written, so the example carries the disable with the reason.

## verification

- `markdownlint-cli2` with the repo config: 0 issues.
- `scripts/verify-harness.sh`: `Level 3 COMPLETE | 0 error(s), 0 warning(s)`, no drift.
- `tests/test_advanced_git_recipes.sh`: 25 assertions, 0 failures.

Came from /retro: yes

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_017yWx1auDfcjcDQDjNg1866)_

### DIFF
--- a/skills/git-workflow/references/advanced-git.md
+++ b/skills/git-workflow/references/advanced-git.md
@@ -476,7 +476,7 @@ git -C <repo> fetch origin && git -C <repo> log --oneline origin/main -3   # or:
 git worktree add ../fresh origin/main                                      # read from a fresh tree
 ```
 
-Two recurring failure modes:
+Three recurring failure modes:
 
 - **Reporting a stale value as fact.** Reading `"^0.13"` from an un-fetched
   worktree and stating "the constraint needs bumping" — when current `main` already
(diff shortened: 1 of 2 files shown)


CybotTM:
## Bot review unavailable — reviewed by hand, and it found something

`pr-status.sh` reports Copilot as **out of monthly review quota**, account-wide: it will not recover this month, on this or any other PR, and re-requesting cannot change it. So this was not merged unreviewed and not merged over a pending review — the review was done by hand, and it is worth saying what it caught rather than just that it happened.

**It found a wrong claim in the section this PR adds.** The text said `set -e` inside `$( )` aborts before the exit code is captured. Reproducing it showed that is only true under POSIX `sh`. Same file, unchanged, three ways:

```text
sh   probe.sh   exit=1   "suite"                    # dies after the header
dash probe.sh   exit=1   "suite"                    # same
bash probe.sh   exit=0   "suite | ok | ok | ende"   # runs clean
```

That split is the load-bearing part and it was missing: a runner written with `#!/usr/bin/env sh` passes while it is developed with `bash runner.sh`, then dies in CI where `sh` is dash or busybox. Fixed in `f7dad03`, together with a note on why a *minimal* probe hides the trap — `bash -c 'set -e; out=$(false; echo $?)'` prints `1` and suggests there is nothing here.

**What else was checked, and how**

- `env VAR=1 my_func` → verified directly: `env: 'f': No such file or directory`, `exit=127`. The claim in the text holds.
- The `--ff-only` reasoning in `advanced-git.md` is standard git semantics and was not re-proved; the claim is definitional, not empirical.
- The exit-127 stale-worktree story is from this session and is what prompted the section.

**Rebase:** the branch was behind `main` after [#229](https://github.com/netresearch/git-workflow-skill/pull/229) landed. Rebased onto `1ab4487`; that PR touched `pull-request-workflow.md` only, so no overlap with the two files here and no conflict.

**Local gates on the final tree:** `markdownlint-cli2` 0 issues · `scripts/verify-harness.sh` Level 3 COMPLETE, 0 errors, no drift · `tests/test_advanced_git_recipes.sh` 25 assertions, 0 failures.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_017yWx1auDfcjcDQDjNg1866)_

PR #236, opened 2026-08-26
fix(hook): let a spent advisory yield, re-arm after six hours

CybotTM:
Two refinements on the once-per-session advisory dedup merged in #226.

**Yield instead of exit.** The already-fired branch used `return`, so a command matching both advisories — a run-poll loop containing a bare git read does — delivered nothing once the first advisory was spent, even while the second still owed the session its first warning. It now `continue`s to the next advisory.

**Side effect, seen and intended.** With `continue`, a command whose only matching advisory is spent also falls through to `check_command`, so its plain-text reminders (force-push, hard-reset) are no longer swallowed by the dedup. The early return when an advisory *fires* stays: one message per call, since mixing the systemMessage JSON with a plain-text reminder on the same stdout would be ambiguous to the harness — both choices are now stated in comments at the spot.

**Re-arm after six hours.** Sessions on the machines this hook serves run for days and their context gets compacted along the way; a warning delivered once at session start is long gone by the time the same slip recurs. Before #226 the named-dir advisory fired 463 times across 54 local sessions, 194 of them in a single session. A marker older than `ADVISORY_REARM_SECONDS` (six hours) now re-arms its advisory: one fresh warning, marker touched, quiet for another window — at most ~4 messages per day instead of 194, while the reminder survives compaction. Every stat/utime failure keeps reading as "warn", same failure posture as #226; a marker mtime in the future (backwards clock step) reads as fresh and stays quiet.

**Tests.** Two new cases, 52 total, all green. Both were watched failing before the fix, and an adversarial review round additionally mutation-verified them: `continue`→`return` reddens exactly the yield test, dropping the `os.utime` touch reddens exactly the re-arm test's third assertion. `ruff@0.16.0` check and format (the versions Skill Validation pins) are clean.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_01C7S9rbgu5giqCwnzwafrHA)_

### DIFF
--- a/scripts/test_validate_git_command.py
+++ b/scripts/test_validate_git_command.py
@@ -13,6 +13,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import time
 import unittest
 from pathlib import Path
 from typing import ClassVar
@@ -285,6 +286,10 @@ class AdvisoryOncePerSession(unittest.TestCase):
     """
 
     ADVISORY_CMD = "git status"  # matches GIT_MEASURING_READ without a -C/cd
+    POLL_CMD = "while true; do gh run list --limit 5; sleep 30; done"
+    BOTH_CMD = (
+        "while true; do gh run list --limit 5; git log --oneline -1; sleep 30; done"
+    )
 
     def setUp(self) -> None:
         self.runtime = tempfile.TemporaryDirectory()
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: 30943d740bdbf1034c6bde71c581f13e9087ca61

PR #238, opened 2026-08-26
fix(pr-status): make --watch usable on a draft PR

### LINKED ISSUE #228: pr-status.sh --watch unusable on a draft PR: NEXT emits "ready", --ignore-action rejects it
`pr-status.sh --watch` cannot be used on a draft PR: `NEXT` emits the action `ready`, but `--ignore-action` refuses it.

## What happens

On a draft PR, `--watch` prints `waiting: draft` once per poll interval — the heartbeat at `scripts/pr-status.sh:1037`, not a return; the loop itself behaves correctly. The problem is that the heartbeat cannot be silenced, because naming the action it reports is refused:

```
$ pr-status.sh -R phpDocumentor/guides 1345 --watch --ignore-action ready
pr-status: --ignore-action: unknown action 'ready'
  (one of: fix-ci triage-ci resolve-threads request-review rebase resolve-conflicts merge blocked none fix-signatures)
```

## Cause

The two vocabularies disagree. `ready` is emitted as a `NEXT` action but is not in the allowlist `--ignore-action` validates against:

- `scripts/pr-status.sh:587` — `{action:"ready", why:"draft", cmd:"gh pr ready …"}`
- `scripts/pr-status.sh:136` — `ACTIONABLE="fix-ci triage-ci resolve-threads request-review rebase resolve-conflicts merge blocked none fix-signatures"`
- `scripts/pr-status.sh:159` — rejects anything outside `ACTIONABLE`

Any action the script can emit should be namable in `--ignore-action`; `ready` is the one that is not. Worth checking whether the allowlist is complete against every `action:` literal in the jq document rather than adding just this one.

## Why it matters now

`references/pull-request-workflow.md` gained the rule (#227) that a PR goes back to draft whenever work resumes on it. Following that rule and then watching the same PR are mutually exclusive today: the state the convention asks for is the state the watcher refuses to hold through. The two features shipped a day apart and collide.

This bites hardest where each stdout line becomes its own notification, as under a line-per-event monitor: a draft PR then produces one notification per interval for a state the operator chose on purpose, and `--ignore-action` — the documented escape for a standing action (#165) — will not take that action's own name.

The workaround is a hand-rolled poll, which `hooks/` correctly denies and which is worse anyway — a loop waiting for `pending == 0` sleeps through a review, which is the failure `--watch` exists to prevent.

## Observed

phpDocumentor/guides#1345 on 2026-08-25, at `pr-status.sh` 1.28.1: PR deliberately in draft during a rebase and a review fix, 31 checks pending, no way to arm a watch for the first actionable event.

CybotTM:
Closes #228.

**The issue's proposed fix was not enough, so this goes one step further.** Adding `ready` to the `--ignore-action` allowlist (the issue's ask) would have made the flag accepted but the watch still useless: `draft` sat near the top of the NEXT ladder, so on a parked PR every poll answered `ready` and a red check, an open thread, or a conflict could never surface — with or without the flag. The ladder placement is the actual defect.

**Change.** `draft` now ranks below the real-work branches (conflicts, stale base, red checks, open threads) and above the review/merge advice, mirroring the queue-entry placement and its rationale. Checks actually running report `wait` — the watch holds and returns on the first real event. Everything else answers `ready`, now in ACTIONABLE: the watch returns handing over `gh pr ready`, and `--ignore-action ready` takes the name (on a settled draft that return is labelled `SETTLED` per the #165 semantics). This makes the back-to-draft-on-resume convention (#227) and `--watch` compatible; the reference now says so.

**Review round caught a regression in the first cut**, verified by the reviewer with a stubbed probe: gating `ready` on `checks_settled` left it unreachable for a draft with zero registered check contexts or a never-dispatched required context (workflows that skip drafts or trigger on `ready_for_review`, fork runs awaiting approval) — an endless `wait` of the very shape this PR removes. `ready` is now gated only on nothing actually running, and the why names undispatched required contexts when present.

**`investigate` joins ACTIONABLE too** — same family: it is the terminal check-branch-protection-manually verdict on settled checks, so a watch heartbeating on it idles into the timeout on a state that waiting cannot change. Observed on #226, where classic-protection CodeQL contexts kept `mergeState=BLOCKED` through an hour of `waiting:` lines. `rules-unavailable` stays a heartbeat on purpose: an API blip can heal on the next poll, and `--max-wait` bounds the true-failure case. Known trade-off (reviewer-flagged, accepted): right after the last check greens, one poll can catch `mergeStateStatus` still at BLOCKED and return `investigate` where the next poll would say `merge` — a cheap exit-and-re-arm, preferred over the unbounded heartbeat.

**Tests.** `tests/test_pr_status_draft_watch.sh` (stubbed `gh`, no network) pins the ladder — hold on running checks; return on a thread, a red check, on `ready`; `ready` reachable with zero registered checks and with a ghost required context; plain status names a red check instead of `ready` — and pins the action vocabulary: every `action:"…"` literal must be ACTIONABLE or a declared waiting heartbeat, and every ACTIONABLE entry must actually be emitted. Watched failing against the respective pre-fix states: 9 assertions red on the original script, 5 more on the version before the `checks_settled` correction. Full `tests/*.sh` suite green, shellcheck clean.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_01C7S9rbgu5giqCwnzwafrHA)_


CybotTM:
Self-review: 3320d1333b9b2cc259256baa1070ec3e690e48da

PR #239, opened 2026-08-26
docs(pull-request-workflow): drop the unverifiable queue-timeout default

### LINKED ISSUE #233: pull-request-workflow.md: verify the claimed check_response_timeout_minutes "default 5"
An adversarial review of #232 pulled on the pre-existing claim in `references/pull-request-workflow.md` that a merge queue drops entries whose checks do not report within `check_response_timeout_minutes` "(default 5)". The reviewer could not confirm that default: the GitHub REST docs specify no default for the parameter (it is required in the ruleset), and the two netresearch repos checked configure 30 (ofelia) and 60 (ldap-manager). If "5" comes from the UI's pre-filled value or an older API version, the reference should say which; if it is unverifiable, the parenthetical should be dropped in favor of "read it from the ruleset" (the section already shows that query).

Out of scope for #232 (the line predates it); filing so the claim gets its own verification pass.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_01D41i7TcscCHnuQ22AJzr4S)_

CybotTM:
Closes #233. The REST docs describe `check_response_timeout_minutes` without stating any default (verified against the current rules endpoint documentation), and the two netresearch repos checked configure 30 (ofelia) and 60 (ldap-manager) — both re-queried live during review — so the "(default 5)" parenthetical asserted a number nothing backs. The paragraph now says the docs state no default and points at the ruleset query the same section already shows.

The review round found the same claim once more: the `pr-status.sh` comment justifying the 5-minute `await-capacity` threshold attributed it to "the tolerance the merge queue itself applies (check_response_timeout_minutes, default 5)". The threshold and behavior are unchanged; the comment now owns the 5 as this script's conservative freshness floor and names the ruleset as the place the real tolerance lives.

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -1967,8 +1967,11 @@ Arm auto-merge / enqueue **only when all four hold**:
 4. **A quiet runner pool** — see below.
 
 **Do not enqueue into a busy runner pool.** A merge queue drops an entry whose
-required checks do not report within `check_response_timeout_minutes` (default
-5). That clock starts at enqueue and covers the wait for a *runner*, not just
+required checks do not report within `check_response_timeout_minutes`. The REST
+docs state no default for that parameter — read the configured value from the
+ruleset (the query below shows it; the two netresearch repos checked run 30
+and 60 minutes).
+That clock starts at enqueue and covers the wait for a *runner*, not just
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: a9f4f00a90c8be51dd716b6d23f75a3e5a13fc30

PR #246, opened 2026-08-27
docs(stash): document recovering a popped/dropped stash via its SHA

aseemann:
## Summary

Document that a popped/dropped git stash's commit object survives (as a dangling commit, reachable by SHA) until `git gc` runs, and how to recover it with `git stash apply <sha>`.

## Came from

`/retro` session on 2026-08-27.
Finding: recover a dropped/popped git stash via its commit SHA
Learning-Id: retro-20260827-recover-dropped-stash-sha

- **Symptom:** After a stash was popped and its changes confirmed restored, an unrelated build-tool hook later silently reverted the same files. The changes appeared lost.
- **Cause:** The agent didn't know a popped/dropped stash's commit object is still recoverable by SHA until garbage collection.
- **Required behavior:** Keep the SHA git prints on stash pop/drop in view; if files thought to be safe turn out reverted, try `git stash apply <sha>` (or `git fsck --unreachable`) before concluding the work is lost.
- **Verification:** Manual — recovered the exact incident's changes via `git stash apply 14386d701aad44499e677dbc4b3a3f613bb6405f` after the stash had already been dropped from `git stash list`.

## Change

Extended the existing "Recovery Operations" section in `skills/git-workflow/references/advanced-git.md` (right after the existing "Recover stash" `git fsck` recipe) with a new subsection, "A popped/dropped stash is not immediately gone," covering: the underlying fact (dangling commit survives until gc), the recovery command (`git stash apply <sha>`), and the practical trigger to watch for (files silently reverted by something other than git after a stash was already popped/restored).

## Test plan

- [ ] Reviewed the new section renders correctly (no broken markdown/code fences)
- [ ] No other file in the repo now contradicts this guidance

### DIFF
--- a/skills/git-workflow/references/advanced-git.md
+++ b/skills/git-workflow/references/advanced-git.md
@@ -1037,7 +1037,9 @@ git gc
 # Aggressive gc
 git gc --aggressive
 
-# Prune unreachable objects
+# Prune unreachable objects — unlike plain `git gc`, this deletes them
+# immediately (no gc.pruneExpire grace period). Don't run it while recovering
+# a dropped stash — see "A popped/dropped stash is not immediately gone" below.
 git prune
 
 # Verify repository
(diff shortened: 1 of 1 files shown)


aseemann:
Self-review: 4130eab9416624eac104231515a5f093ca1bac56

PR #252, opened 2026-08-28
feat(pr-status): report issue comments nobody answered

CybotTM:
`pr-status.sh` reports two feedback channels — the check rollup and `reviewThreads` — and treats their silence as silence overall. There is a third. A reviewer who writes prose *under* the pull request rather than on a line of the diff produces an **issue comment**, which appears in neither, so the tool prints `threads: 0 unresolved` while the findings sit unread and the `NEXT:` line advances to `request-review` or `merge`.

I hit this on [phpDocumentor/guides#1352](https://github.com/phpDocumentor/guides/pull/1352) on 2026-08-28: three comments waiting, two of them substantive findings from a maintainer (an untested option combination, and a regression-test value that missed `PHP_INT_MAX` by seven orders of magnitude). I reported the PR as having nothing open, on the strength of this tool, until a human asked whether I had read them.

## What changed

The GraphQL query already fetched the comments — they were used only to find the author's own `Self-review:` attestation. This adds `createdAt` to that field, counts what arrived after the last word of the author, reports it, and adds an `address-comments` rung to the ladder.

Placement is deliberate: **below** `resolve-threads`, which names a file and a thread id and is the more specific answer, and **above** draft and every review and merge branch, because unread feedback is real work however the PR is parked.

Bot comments are counted and shown but never drive the ladder (`$unanswered_human`), so a Renovate or Dependabot note cannot push its own PR off the auto-merge rung it exists to reach. The signal for that is the GraphQL `__typename`, not the login: GraphQL returns `Bot` for an App and strips the `[bot]` suffix REST appends, so a login test alone reads `github-actions` and `sonarqubecloud` as people. The first commit here got that wrong and **this pull request demonstrated it on itself** — its own CI comments produced `NEXT: address-comments` for two machine notices. Fixed in the second commit, with a case pinning it.

## What it looks like

```
  threads     : 0 unresolved
  comments    : 1 unanswered (linawolf)

NEXT: address-comments — 1 comment(s) posted after your last word by linawolf — issue comments live outside reviewThreads, so "0 unresolved" above does not cover them
```

## Verification

`tests/test_pr_status_unanswered_comments.sh` drives the real script against a stubbed `gh`, six cases: a maintainer comment newer than the author's last word, the author having answered afterwards, a bot-only comment, an App whose login carries no `[bot]` suffix, an author who never commented at all, and an open review thread that must stay the more specific answer. Without this change the test fails, and the failure is the defect itself — `NEXT=merge` for a PR carrying an unread maintainer comment.

The repo's own suite caught two things I had missed on the first pass and both are fixed here: the four new `--json` keys were absent from the `--json contract` header (`test_pr_status_json_contract.sh`), and `address-comments` was emitted without being classified as actionable (`test_pr_status_draft_watch.sh`), which would have made `--watch` sleep through it.

Full suite (13 files), `scripts/verify-harness.sh` and `shellcheck` are green. Also ran the patched script against eight live PRs on phpDocumentor/guides: the new line appears on exactly the three whose newest comment is from someone other than the author, and on none of the other five.

No version bump — that belongs to the release flow.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Mf63edGvCVRQz6mwF8gxcC)_

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -31,10 +31,18 @@ bash "$GW/scripts/pr-status.sh" -R owner/repo 123 --watch
 Two API calls (a third, admin-only, when the PR is review-blocked — it reads
 classic branch protection, whose review gates like `require_last_push_approval`
 are invisible to the rules endpoint), and the output ends in a computed `NEXT:` — rebase, fix-ci,
-triage-ci, resolve-threads, request-review, wait, or merge (with the method
+triage-ci, resolve-threads, address-comments, request-review, wait, or merge (with the method
 this repo actually allows and a warning when a merge queue is active). The
 JSON form carries each unresolved thread's `threadId` *and* `commentId`, which
 is everything needed to reply and resolve without another query.
+
+`address-comments` covers the channel the other two miss. A reviewer who writes
+under the pull request instead of on a line of the diff produces an *issue
+comment*, which lives in neither `reviewThreads` nor the check rollup — so a
+report built from those alone says `threads: 0 unresolved` while the findings
+sit unread. Anything posted after the last word of the author counts as
+unanswered; comments from bots are shown but never drive the `NEXT:` line, so a
+Renovate note cannot push its own pull request off the auto-merge rung.
 
 Measured on a 40-PR rollout that did not have it: **183 of 370 shell calls
 were PR-status probing**, the rulesets endpoint was queried exactly once, and
--- a/skills/git-workflow/scripts/pr-status.sh
+++ b/skills/git-workflow/scripts/pr-status.sh
@@ -76,6 +76,7 @@
 #   state mergeable mergeState draft number title repo author
 #   base head headOid
 #   checks checks_settled threads unresolved_threads
+#   unanswered_comments unanswered_human unanswered_by unanswered_urls
 #   reviewDecision reviews_on_head has_review_on_head
 #   has_copilot_review_on_head copilot_review_errored copilot_error_count copilot_quota_hit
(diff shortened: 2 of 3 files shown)


CybotTM:
Copilot is out of review quota for this month, account-wide, so no bot review is available on this head and re-requesting cannot change that. Recording what I checked instead.

Read the full diff of both commits. The change adds no API call — `comments` was already fetched for the `Self-review` attestation and only gained `createdAt` and `__typename`. Timestamps are compared as strings, which holds because GitHub returns fixed-width UTC (`…Z`); an offset-bearing format would break it, and that is the one assumption in here worth knowing about.

Ladder placement exercised against the verbatim expression with a mutated state: the new rung fires on a human comment, yields to `resolve-threads` and to `fix-ci`, and is skipped when the count is zero. Six stubbed cases in `tests/test_pr_status_unanswered_comments.sh` pin it; without the change the test is red, and the red is the defect itself (`NEXT=merge` for a PR carrying an unread maintainer comment).

Ran the patched script against eight live PRs on phpDocumentor/guides: the new line appears on exactly the three whose newest comment is by someone other than the author. The first commit then failed on this very PR — its own CI comments read as people — which is how the `__typename` fix in the second commit was found rather than assumed.

Full suite (13 files), `scripts/verify-harness.sh` and `shellcheck` green. SonarQube quality gate passed; its 18 new findings are all `S7679`/`S7688` style rules on the new test file, matching the `check()` helper every one of the ten existing test files uses — 314 of the same findings are open repo-wide, so this file follows the convention rather than departing from it.

No human has read this change.

Self-review: c7dc5a3325e0d3d106c994a104837da0bff0d725

PR #277, opened 2026-09-04
docs(commit-conventions): a fix that removes behaviour is at least MINOR

CybotTM:
The type table maps the *intent* of a change. The version has to describe what a **consumer** observes, and the two come apart in one recurring case: a `fix:` that takes away something the consumer was getting before.

> If a change takes away something the consumer was getting before, it is at least MINOR — even when the commit is a `fix:`.

`fix:` says *why* you are doing it. It does not say that nobody depended on the behaviour you removed.

## The case it came from

Kept in the text, because a rule without its case reads as an opinion.

An Ansible role pulled unattended-upgrades from **every** APT origin, third-party repositories included — so it restarted `dockerd` on an ingress node and on both halves of a DNS resolver pair, outside any maintenance window. Scoping it to the distribution sources was genuinely a defect fix: the sibling platform had been scoped correctly all along. It was tagged as a patch.

But hosts that had been getting third-party packages automatically stopped getting them, and nothing in the version said so. A consumer upgrading a patch level does not read release notes — that is the entire point of patch levels.

Both of us reached for PATCH first, for the same reason: the commit was a `fix:`, and the change repaired a real defect. The question that settles it is not "is this a fix" but "does the consumer get less than before".

## Also recorded

What to do when the patch tag is already published: **do not re-tag the same commit.** Two tags on one commit create two truths for one change, which is worse in hindsight than a version one step too low. Put the behaviour change where the next reader will look — the tag message, the release notes — and move on.

`BREAKING CHANGE:` keeps its meaning: what breaks a consumer outright. This rule covers the quieter case in between — nothing breaks, something stops.

## Gates

`pre-commit run --files skills/git-workflow/references/commit-conventions.md` — markdownlint, trailing whitespace, end-of-file, all clean.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01LxFXkREDeH7z1pjcnn5uAF)_

### DIFF
--- a/skills/git-workflow/references/commit-conventions.md
+++ b/skills/git-workflow/references/commit-conventions.md
@@ -28,6 +28,44 @@
 | `chore` | Maintenance tasks | - |
 | `revert` | Reverting changes | - |
 
+### The type is not the whole answer for the version
+
+The table maps the *intent* of a change. The version has to describe what a
+**consumer** observes, and the two come apart in one recurring case:
+
+> **If a change takes away something the consumer was getting before, it is at
+> least MINOR — even when the commit is a `fix:`.**
+
+`fix:` says *why* you are doing it. It does not say that nobody depended on
+the behaviour you removed. Fixing a defect that was quietly doing someone a
+favour still changes what they get.
+
+A worked case, because a rule without the case it came from reads as an
+opinion:
+
+```
+fix(apt): never upgrade docker-ce or containerd.io unattended
+```
+
+Genuinely a defect — the role pulled from every APT origin including
+third-party repositories, so unattended-upgrades restarted `dockerd` on an
+ingress node and on both halves of a DNS resolver pair, outside any window.
+The sibling platform had been scoped correctly all along, so this was
+repairing an oversight, and it was tagged as a PATCH.
+
+But hosts that had been getting third-party packages automatically stopped
+getting them. Nothing in the version said so. A consumer upgrading a patch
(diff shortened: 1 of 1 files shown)


CybotTM:
**Copilot review unavailable — self-review below.**

`copilot_code_review` returned a quota error on this head; the wall is account-wide, so re-requesting cannot clear it. Recording what I checked instead of merging silently.

**What the change is.** One section added to `references/commit-conventions.md`: that a change removing behaviour a consumer was getting is at least MINOR, even when the commit is a `fix:`. Documentation only — no script, no gate, no test touched.

**What I checked.** The section is internally consistent with the rest of the file: it does not contradict the type table above it, and it keeps `BREAKING CHANGE:` reserved for outright breakage, positioning itself as the quieter case in between. The worked example is concrete and carries its own reasoning — an unattended-upgrade fix that stopped third-party packages arriving on hosts that had been getting them — which is what makes the rule checkable rather than an opinion. Its advice against re-tagging an already-published patch is sound and matches the release practice used here: two tags on one commit are worse than a version one step too low.

**What I did not do.** I did not verify the incident behind the example against its source repository. It reads as first-hand and nothing in it is self-contradictory, but I am not attesting to that history — only that the rule the section teaches is coherent and correctly scoped.

The branch was rebased onto current `main` before this; the content is unchanged by that.

CybotTM:
Self-review: 199c48aa40ac05f4f73cfc8d1a8731c1d4d04bc0

PR #282, opened 2026-09-09
docs(pull-request-workflow): stop reading an empty requested_reviewers as the quota tell

CybotTM:
Two references in this skill gave opposite readings of the same measurement, and following the wrong one produced a false statement in a live session.

`references/merge-gate-watcher.md` says to confirm a review request off the timeline, "not off `requested_reviewers`: the response body does not echo a bot reviewer, and a reviewer that has started drops off that list without having submitted". `references/pull-request-workflow.md` said the opposite — that an empty read-back right after the POST "is the earliest quota tell there is", and that empty means stop requesting everywhere.

## What it cost

Requesting a Copilot review on netresearch/git-workflow-skill#281, the read-back came back empty. Following the second instruction I reported the request as swallowed by the quota. It had not been: `pr-status.sh` showed a minute later that Copilot had answered on that very pull request, with an error body naming the quota limit — `copilot_quota_hit` was true off the review row, not off the empty array.

## The edit

The observation from 2026-08-18 stays — a POST really can be dropped, and that is a genuine early face of the wall. The instruction derived from it does not, because the array has three producers and cannot tell them apart: a bot reviewer is never echoed at all, a reviewer that has started has already left the list, and only the third case is a dropped request. The paragraph now says that and points at the timeline check the other reference already prescribes. No third instruction is added on top of the two; the stale side is superseded.

## Also recorded

One bot account reaches callers under three logins, all naming the same Renovate install and measured on netresearch/github-release-skill#110: GraphQL answers `renovate` with `__typename: Bot`, `gh pr view --json author` answers `app/renovate`, and the webhook payload — what `auto-merge-deps.yml` matches on — answers `renovate[bot]`. A check against any one hardcoded name is wrong for the other two. This lived only in a code comment in `pr-status.sh` after #281; anyone writing a bot check elsewhere would re-derive it.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Utz6KTyv1TNKMvXpqyNSv5)_

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -183,14 +183,21 @@ reset rather than being aged out; delete the file to undo a verdict recorded in
 error.
 
 The quota wall has a second, earlier face: the `requested_reviewers` POST
-itself can be **silently dropped** — HTTP 200, but the response's
-`requested_reviewers` array is empty and a read-back seconds later still shows
-no pending request, with no errored review anywhere yet. That empty read-back
-right after the POST is the earliest quota tell there is; it looks like "the
-bot will pick it up asynchronously" and is not (five requests across four repos
-were swallowed this way on 2026-08-18 before a later errored review named the
-quota). Read `requested_reviewers` back after the **first** request of a
-session; empty means stop requesting — everywhere — and go straight to the
+itself can be **silently dropped** — HTTP 200, but no pending request appears
+and no errored review follows either (five requests across four repos were
+swallowed this way on 2026-08-18 before a later errored review named the
+quota).
+
+An empty `requested_reviewers` does **not** establish that, and reading it as
+the quota tell states a swallowed request where there was none. The array has
+three producers and cannot tell them apart: the response body never echoes a
+**bot** reviewer at all, a reviewer that has **started** drops off the list
+without having submitted, and only the third case is a genuinely dropped POST.
+Confirm off the **timeline** or the reviews list instead, as
+`merge-gate-watcher.md` § *"What still works while GraphQL is drained"*
+prescribes — an errored Copilot row on the head is the same wall, said out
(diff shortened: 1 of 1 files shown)


CybotTM:
Both review bots are unavailable on this head, for different reasons. Copilot answered with a quota error — the wall is recorded account-wide for 2026-09 and `pr-status.sh` reports `copilot_quota_hit`. CodeRabbit declined rather than reviewed: "You've used all free OSS reviews for now", next one in 59 minutes. Per the documented fallback I reviewed the diff myself; this records what was looked at and where the evidence for each claim comes from.

Verified in this session: that an empty `requested_reviewers` cannot establish a swallowed request. Requesting a Copilot review on #281, the read-back was empty while Copilot had in fact already answered — the "reviewer that has started drops off the list" producer, observed directly. The bot-login paragraph is measured too, on netresearch/github-release-skill#110, all three forms read back from the live API.

Not verified by me, and stated as such: that the response body never echoes a bot reviewer at all, and the 2026-08-18 record of five genuinely swallowed POSTs. Both come from the skill's own references — the first from `merge-gate-watcher.md`, which this change sides with, the second from the paragraph being edited. The operative instruction does not rest on either: whichever of the three producers applies, the array cannot tell them apart, so the timeline is the read that answers.

Also checked: the cross-reference names the real heading in `merge-gate-watcher.md` (§ "What still works while GraphQL is drained") rather than one I invented, and `copilot_quota_hit` is a field `pr-status.sh` actually emits — case 7c of its suite asserts it. Markdownlint and the 23 checks are green.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Utz6KTyv1TNKMvXpqyNSv5)_

CybotTM:
Self-review: 435a92c301df782ed61f9409fd94e7542abf7f6e

PR #283, opened 2026-09-09
docs(pr-workflow): CodeRabbit declines out loud and does not catch up

CybotTM:
From a `/retro` on a session that lost about an hour to bot-review behaviour this reference did not describe.

## The pending-review half of the force-push rule

The reference already says a force-push invalidates a **completed** review. It does not say the same push discards a review that was *requested and not yet delivered* — and that half is worse, because nothing reports it.

What happened: three reviews requested, each acknowledged by the bot, then the branches rebased onto a merged sibling. The force-push threw away all three runs. The PRs sat `CLEAN` with `reviews: NONE on current head` and no signal distinguishing "never requested" from "requested and discarded". Hence the ordering rule now stated: **rebase first, then request.**

## Two CodeRabbit properties that decide whether waiting means anything

- **It reviews on events, not on backlog.** Its own wording: *"CodeRabbit is an incremental review system and does not re-review already reviewed commits."* A review that did not happen at push or ready-for-review time is not pending — it is not coming.
- **A rate limit is an answer, not silence.** On the free OSS tier the reply to `@coderabbitai review` is a comment reading `Review rate limited.` under an **`⚠️ Action not completed`** heading. Counting `reviews: []` cannot tell that apart from a review still running.

Together those make a wait loop keyed on "a review will appear" able to spin forever — which is exactly what happened before the refusal comment was actually read.

- **A clean review leaves no review object either.** Found by walking into it on this PR, minutes after writing the two above. CodeRabbit reviewed the head, found nothing, and posted only `No actionable comments were generated in the recent review.` — no review submitted, no inline comments, `reviews[]` empty, `pr-status.sh` still reporting `reviews: NONE on current head`.

So one empty array covers three terminal states: refused, never triggered, and reviewed-with-nothing-to-say. The section now gives the one-line query that separates them, and notes that the clean case names the commit range it covered — so it can be checked against the current head instead of assumed.

## The both-walled case

Copilot out of monthly quota **and** CodeRabbit rate limited is reachable, and was reached. The section now names it and points at the attestation fallback the tooling already implements, instead of leaving the operator to invent something.

It is worth doing properly rather than treating it as a formality: in that same session the one PR CodeRabbit did review returned a real defect, and reading the remaining three by hand found four more — including one that would have failed CI in every fleet repository on a correct input.

## Scope

Docs only, one file, one section. No script, schema or `pr-status.sh` behaviour is touched — the tool already reports the Copilot wall; this is the operator-facing half it cannot express in a `NEXT:` line. `markdownlint-cli2` clean; the test suite passes (`test_signing_preflight.sh` references this file in a comment only).

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01L94yEWSBvuUkLqLpVbMqRh)_

### DIFF
--- a/skills/git-workflow/references/pull-request-workflow.md
+++ b/skills/git-workflow/references/pull-request-workflow.md
@@ -137,10 +137,14 @@ Two properties decide whether waiting for CodeRabbit is worth anything:
   is an incremental review system and does not re-review already reviewed
   commits."* A review that did not happen at push or ready-for-review time is not
   pending — it is not going to happen, and waiting produces nothing.
-- **A rate limit is an answer, not silence.** On the free OSS tier the reply to
-  `@coderabbitai review` is a comment reading `Review rate limited.` under an
-  **`⚠️ Action not completed`** heading. Read it: a request that was refused looks
-  identical to one still running if you only count `reviews: []`.
+- **A rate limit is an answer, not silence.** The reply to `@coderabbitai review`
+  is a comment reading `Review rate limited.` under an **`⚠️ Action not
+  completed`** heading. Read it: a request that was refused looks identical to
+  one still running if you only count `reviews: []`. The allowance is hourly and
+  applies to public repositories independently of the organisation's plan — the
+  same notice reports `Plan: Advanced`, `up to 1 included review per hour` and
+  "you've used all free OSS reviews for now" together, so a paid plan is no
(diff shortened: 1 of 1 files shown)


CybotTM:
Self-review: 37da8c323fd4979905613107a07ed761a6295944

PR #313, opened 2026-09-17
fix(pr-status): a review is mandatory, a bot review is not

CybotTM:
Merging this lets a reviewed pull request merge in a repository that has no `copilot_code_review` ruleset, which today it cannot. A review stays mandatory; what stops being mandatory is that a *bot* wrote it.

The self-review attestation counted only where a Copilot review was unsatisfiable — a quota wall, or two failures on the head. That tied the whole self-review path to a bot most repositories neither require nor have the quota for. In a repository without the ruleset the generic refusal stamped no reason at all, so `--self-reviewed` answered "a live review path exists, use it" while Copilot was out of monthly quota and CodeRabbit replied `Review rate limited` — a live path that did not exist. Found on [netresearch/agent-system-evals#85](https://github.com/netresearch/agent-system-evals/pull/85): green, no open threads, a written review in the PR, and no way to merge it.

**The rule this encodes.** A review is mandatory and may be the author's own. A bot review is waited for only while one is actually in flight — requesting a reviewer commits you to waiting for its answer, so a pending Copilot request keeps the attestation inert and the ladder answers `await-review`. Nothing else waits, because nothing else is coming.

- `$self_attested` drops `copilot_quota_exhausted or copilot_exhausted` and gains a guard on a pending Copilot request. The `CHANGES_REQUESTED` and `REVIEW_REQUIRED` guards are untouched: a human saying no, and a host demanding an approval, are not demands an attestation can satisfy.
- The generic branch stamps `next.reason: review-required`; the ruleset branch stamps it too, since its own comment already says the demand there is tool policy rather than a host gate. `pr-merge.sh` accepts that reason alongside `bot-review-unsatisfiable`.
- Both the `NEXT: merge` line and the attestation body now name *which* demand is satisfied. Writing "unsatisfiable" where the truth is "a review was required and I wrote it" puts a false claim about a bot into permanent pull-request history — the same reason the reason-keying exists at all.

**Tests.** The case that asserted the old rule (`SR4: a live review path wins`) is flipped rather than deleted, and four join it: an in-flight Copilot request, a host-required approval, the generic demand naming its reason, and `pr-merge` accepting it with a body that does not claim an unsatisfiable bot. Checked against the old scripts rather than assumed: three assertions in `pr-status` and three in `pr-merge` fail there; the two guard cases pass both ways, because those guards already existed. Full suite green, validator reports 0 errors.

`SKILL.md` and `references/pull-request-workflow.md` are updated in the same commit — a tool whose manual describes the rule it no longer implements is the defect I spent this morning fixing elsewhere.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_


CybotTM:
Both bots are walled on this head — CodeRabbit refused it as rate limited, Copilot is out of monthly quota — so I reviewed the diff myself. Worth noting that this pull request is the fix for the case where that leaves no path at all; here the older `bot-review-unsatisfiable` reason still applies, because this repository has the `copilot_code_review` ruleset.

What I checked rather than argued:

The behaviour change is four assertions wide and I ran each against the old scripts as well as the new. `SR4` (attestation, no wall, no request in flight) goes from `request-review` to `merge`; `SR4d` (no attestation) now stamps `review-required` where it stamped nothing; `pr-merge` case 2b goes from refusal to a posted attestation whose body says "requires a review and no bot review is in flight". Three assertions in `pr-status` and three in `pr-merge` fail against the old code. `SR4b` and `SR4c` pass both ways, which is the point: the in-flight guard and the `REVIEW_REQUIRED` guard are not new, and this change must not weaken them.

The two guards that carry the safety of the whole change are the ones I read most carefully. `REVIEW_REQUIRED` keeps the attestation inert because a host-required approval is not something an assertion can supply; `CHANGES_REQUESTED` keeps it inert because a human said no. Neither is touched. The new guard — a pending Copilot request — is what keeps "a review is mandatory" from collapsing into "never wait for a bot".

One judgement call: the ruleset branch now stamps `review-required` too. Its own comment already said the demand there is the never-merge-unreviewed policy rather than a host gate, and a Copilot review counts toward no required approval, so an author review satisfies the same demand. If that reading is wrong, the fix is to drop the reason from that branch alone; the generic branch is the one this pull request exists for.

Full suite green, validator 0 errors, and the manual is updated in the same commit — a tool whose documentation describes a rule it no longer implements is the defect I spent this morning fixing elsewhere.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_

CybotTM:
Self-review: de41b8ae8df0e741b0383fbfcb90b5ffcd4301b5

PR #315, opened 2026-09-18
docs(pull-request-workflow): range-diff is what says a review survived a rebase

CybotTM:
Merging this gives the skill an answer to "the rebase invalidated my review, do I have to read the whole diff again?". `git range-diff` appeared nowhere in the skill before this.

## The gap

The self-review attestation dies with the next push, and a rebase is a push — the file says so already, two paragraphs above where this lands. Under `strict_required_status_checks_policy: true` that is not an edge case: every merge into the base makes the remaining sibling PRs stale, so a sweep rebases each one in turn and arrives at a fresh head carrying no review, while the review that *was* done still describes the code exactly.

The skill had no way to establish that second half. The choices it left were re-reading the whole diff, or attesting on a feeling.

## What this adds

```bash
git range-diff origin/main...<old-head> origin/main...<new-head>
```

Commits marked `=` are byte-identical in their patch; a `<`/`>` pair carries a diff-of-diffs and is the commit that needs reading. An all-`=` series means the rebase moved the base and nothing else, so an earlier review — the author's, a bot's, a human's — still applies to the new head and can be named as the basis for the attestation.

## Why it is not the check already documented

`rebase_long_branch_with_reference_merge` and its section cover the neighbouring question: build the intended end state in a scratch worktree and `git diff` the rebased branch against it, which must print nothing. That proves the resulting **tree** is right. `range-diff` proves the **patches** are still the ones somebody read. A rebase can reach the correct tree through commits that changed on the way, and reviewers read commits, so the two are not substitutes. Both are now in the file, each labelled with the question it answers.

The section also notes the case that prompted it: CodeRabbit and Copilot both refuse heads while rate-limited or out of quota, and a refusal is not a review. Whether the review you already hold still describes this code is the part you can settle yourself.

## Changes

- `references/pull-request-workflow.md`: new subsection after *Putting the self-review on the record (#203)*, where the attestation's push-invalidation is already stated.
- `evals/evals.json`: `rebase_invalidated_the_review`, asserting the answer names `range-diff`, the three-dot form, the meaning of `=`, and that a differing commit gets read.

`validate-evals.sh` from skill-repo-skill: 135 passed, 0 failed, against 133 on `main` at `42860b3` — the two added checks are the new eval's prompt and its four assertions. Pre-commit hooks passed (markdownlint, check json, ruff format). Rebased onto `main` after #314 merged; `git range-diff` reports the commit as `=`, so the patch is unchanged and only the base moved.

## Not included

No change to `SKILL.md`. Its `Content Triggers` table already routes "PR merge gate, signed rebase" to this file, and the body is against a 500-word gate where spending words has to buy a capability that would otherwise be invisible — this one is inside a file the table already points at.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Gswmp3AQngaYiJ4GuHSUSY)_


CybotTM:
Self-review: 45a0a88ce479dcad10ae6e9035e646e31cc458db

PR #318, opened 2026-09-18
docs: three PR and revert facts a session had to re-derive

CybotTM:
Merging this adds three paragraphs to the skill's references — two in `pull-request-workflow.md`, one in `advanced-git.md`. No script, no behaviour and no other file changes.

All three come from a 2026-09-18 session that reverted a merged PR in a foreign repository ([phpDocumentor/guides#1397](https://github.com/phpDocumentor/guides/pull/1397), reverting [#1357](https://github.com/phpDocumentor/guides/pull/1357)) and had to work each one out from scratch. Each absence was verified against `main` with a whitespace-normalised scan of `skills/git-workflow/**/*.md`, not a line grep — prose in this repo is hard-wrapped, so a phrase that reads as one sentence spans two lines and a line grep reports it missing whether it is there or not.

## A revert PR's first fact is whether the break is in a release

`git tag --contains <merge sha>` on the commit being reverted. Empty means the break never shipped, which is both the strongest sentence such a PR has and a deadline the maintainer needs to know about; a tag means naming the version consumers are on and deciding whether the revert has to reach a maintenance branch too. The section writes out the commands — including reading `.backportrc.json`'s target branch and checking whether it exists — and the paragraph each answer produces.

On the session's PR this was asked only on the second pass. The merge was three months newer than the latest tag and the configured backport branch did not exist upstream at all, so two facts that decided the PR's whole argument nearly went unstated.

## A PR's draft state can change from outside your session

The skill already says a ready PR goes back to draft before the first push when work resumes. It does not say how to tell that case from a human deliberately marking it ready — and on a shared account `gh pr view --json isDraft` cannot: the field reads `false` either way and the actor on the event is your own login in both cases. The addition gives the timeline query that separates them and dates the change against your own pushes, and says what to do with the answer: report it, do not silently convert it back.

The existing `ready_for_review` mentions in this file are about workflow triggers, not attribution, so nothing is duplicated.

## Keeping a revert commit pure

In `advanced-git.md`: when a revert arrives together with an improvement, commit one has to stay something a reviewer can verify against `git revert`'s own output. The recipe sets the edit aside and restores the pristine content with `git show HEAD:<old path>`, notes that the old path must be staged alongside the new one or the rename records as an add plus a stray deletion, and says why `git checkout --` is the wrong way back — it restores the last committed state and takes the new edit with it. The scratch copy goes outside the worktree, for the reason the neighbouring `git add -A` section already gives.

## Verification

`bash tests/test_advanced_git_recipes.sh` — 72 assertions, 0 failures. Pre-commit ran on the commit: markdownlint, trailing whitespace, end-of-file and the rest, all passed.

`Build/Scripts/validate-skill.sh` exits 1, and it does so on `main` too: `SKILL.md is 602 words (max 500)`, and `git show origin/main:skills/git-workflow/SKILL.md | wc -w` is also 602. This PR touches no `SKILL.md`. Trimming it is an editorial change that deserves its own PR rather than a drive-by here, but the repo's own validator being red on `main` is worth someone's attention.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01EWiYVcHSD71piDkNv2r8c6)_


CybotTM:
Self-review: 9e6cf3e0423406246041af6c83ea0b49db216b46

PR #321, opened 2026-09-18
docs(skill): route "my watcher got killed" to the file that answers it

CybotTM:
Merging this makes `references/merge-gate-watcher.md` findable from the reference table by someone whose watcher just died — which is the situation the file's own text describes and the table's trigger cell does not mention.

The passage at line 180 reads: *"Prefer the Monitor tool over a backgrounded shell for these long waits. Two backgrounded `pr-status.sh --watch` processes in one session were stopped by the harness with 'the system is running low on memory' … the same watch under Monitor ran to its event."* That is exactly right, and it went unread during a `/pr-finish` run on 2026-09-17 where the same thing happened three times. The trigger cell says `Merge-driver loop, check taxonomy, stale-SHA rerun`; none of those words match "my watcher keeps getting killed", so the table routed the reader nowhere and the workaround got invented instead: a hand-rolled poll loop (correctly refused by the poll gate), then a foreground run with a Bash timeout.

The change is the two missing words in that one cell. The reference itself is untouched, as is every other row.

### Checks

`validate-skill.sh` from the repository root: 0 errors, 20 warnings — unchanged from the base. `markdownlint-cli2` on `SKILL.md`: 0 issues. SKILL.md body is 74 lines, far under the 500-line cap, so no content had to move.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01T3jnqu5aTUeed5LtZugtYx)_

https://claude.ai/code/session_01T3jnqu5aTUeed5LtZugtYx


CybotTM:
Copilot review unavailable — the account's monthly quota is exhausted, and its single row here is a refusal rather than a review ("Copilot was unable to review this pull request because the user who requested the review has reached their quota limit"), recorded against a head that two rebases have since replaced. CodeRabbit posted a walkthrough but never submitted a review on this pull request.

My own read of the diff, since nothing else is coming:

- The claim the new trigger words make is true of the file they point at: `references/merge-gate-watcher.md` does cover a watcher the harness killed, by name and with the instrument that survives it, in the paragraph beginning "Prefer the Monitor tool over a backgrounded shell for these long waits."
- The row keeps the two-column shape of its neighbours and the reference path is unchanged, so nothing that links to this file by path is affected.
- Adding words to a trigger cell can only widen what routes here; it removes no existing route, since the three original terms are all still present.
- `validate-skill.sh` from the repository root reports 0 errors and 20 warnings, the same as the base, and `markdownlint-cli2` on SKILL.md reports 0 issues. The body is 74 lines against the 500-line cap, so nothing had to move into a reference to make room.

Merging on that basis.

CybotTM:
Self-review: ac32fe36d2e3a8e6cc89070a878ed1d6fab8c548

Issue #319, opened 2026-09-18
address-comments never clears on a bot-authored PR, so pr-merge.sh refuses a green, approved, thread-free pull request


CybotTM:
On a bot-authored pull request, `address-comments` can never clear, so `pr-merge.sh` refuses a pull request that is green, approved and thread-free.

## What happens

`pr-status.sh` computes the `address-comments` rung from comments posted after the **author's** last word. On a Renovate or Dependabot pull request the author posts once, at creation, and never again. Every comment that follows therefore stays counted for the life of the pull request — the bot status notes the repository itself produces (Codecov, SonarCloud, Dependency Review, and CodeRabbit's own "Review skipped — bot user detected"), and the reviewing human's comments too, including the approval note the gate's own documented path asks for.

Observed on netresearch/t3x-contexts_geolocation#56, after following `references/pull-request-workflow.md` § *A bot-authored pull request takes the other path (#280)* to the letter:

```
state       : OPEN  mergeable=MERGEABLE  mergeState=CLEAN
checks      : 62 pass, 0 fail, 0 pending, 9 skip (of 71)
reviews     : CybotTM=APPROVED  decision=APPROVED
threads     : 0 unresolved
comments    : 5 unanswered (CybotTM, codecov, coderabbitai, github-actions, sonarqubecloud)

NEXT: address-comments — 1 comment(s) posted after your last word by …
```

```
pr-merge: not merging netresearch/t3x-contexts_geolocation#56 — address-comments: …
```

Every substantive axis is open: `CLEAN`, nothing red, nothing pending, zero unresolved threads, an `APPROVED` review on the current head. Only the comment counter is shut, and no action available to the reviewer closes it: answering the comments adds another comment by me, which the counter also counts, because it is measured against the author's last word and the author is a bot.

## Why the current rule is right for the case it was written for

For a human-authored pull request the rung is exactly correct — a reviewer wrote something under the pull request instead of on a line of the diff, the author has not replied, and the author is the one who should. The `references/pull-request-workflow.md` text says so: *"Anything posted after the last word of the author counts as unanswered"*, and *"comments from bots are shown but never drive the `NEXT:` line, so a Renovate note cannot push its own pull request off the auto-merge rung."* That second guarantee is the one that does not hold here — not because bot comments drive it, but because the human reviewer's own comments do, with no way to answer them into silence.

## Suggested rule

[...]

Issue #280, opened 2026-09-09
pr-merge.sh has no path for a bot-authored PR that policy routes to a human


CybotTM:
A Renovate or Dependabot pull request that a repository deliberately routes to a human has no path through `pr-merge.sh`, so the merge happens outside the script that exists to be the safe path.

## Where it stops

`pr-merge.sh` line 143 requires the authenticated user to be the pull request's author before it will post a `Self-review:` attestation:

```
--self-reviewed refused: the attestation must come from the PR author (renovate); gh is authenticated as CybotTM
```

The check is right — an attestation is an assertion by the person who read the diff, and letting a third party mint one for someone else's pull request would hollow it out. But a bot never authenticates and never reviews, so for a bot-authored pull request the condition can never hold.

Without the flag the script refuses too, because there is no review on the head:

```
pr-merge: not merging …#110 — request-review: no review on the current head (4daf8b61) — do not merge unreviewed.
```

Both refusals are correct in isolation. Together they leave no sanctioned command.

## Why the deps workflow does not cover it

The usual answer — "dependency pull requests auto-merge, never merge them by hand" — does not apply to the pull requests that reach this state. `netresearch/.github`'s `auto-merge-deps.yml` excludes them by label:

```yaml
if: >-
  (github.event.pull_request.user.login == 'dependabot[bot]' ||
   github.event.pull_request.user.login == 'renovate[bot]') &&
  !contains(github.event.pull_request.labels.*.name, 'deps-no-automerge') &&
  !contains(github.event.pull_request.labels.*.name, 'deps-major')
```

and its own input documentation states the intent: *"Majors are approved but left for a human to merge."* So the label means a human merges it — exactly the case `pr-merge.sh` cannot serve. Removing the label does not help either: the caller triggers on `pull_request_target` with default types, which do not include `unlabeled`, and re-running the skipped run re-evaluates against the stored payload that still carries the label.

## What happened in practice

netresearch/github-release-skill#110 (a `deps-no-automerge` major) was verified by hand — the breaking change in the new major was measured against the repository, 16 warnings and 0 errors under both the old and the new version — and then merged with a bare `gh pr merge --merge`, around the script. That is the shape the skill's own reference warns about under "Putting the self-review on the record": a merge running past the very script that exists to be the safe path.

## Worth deciding, not necessarily patching

Options, in the order I would weigh them:

[...]

PR #327, opened 2026-09-18
fix(pr-status): stop advising --self-reviewed to a non-author

CybotTM:
## What changes when this merges

`pr-status.sh` stops recommending `pr-merge.sh --self-reviewed` to an operator who did not author the pull request, and recommends the ordinary `gh pr review --approve` instead. The self-review attestation is an assertion *by the author* and `pr-merge.sh --self-reviewed` refuses every other authenticated user, so that advice could never succeed — it was reaching non-authors because the decision was made from `$author_is_bot` alone.

## How it was found

On netresearch/concourse-ci-skill#57 — author `aseemann`, viewer `CybotTM`, both human. Every status read and every `--watch` tick printed the attestation advice, it was relayed to the operator twice as the way forward, and `pr-merge.sh` refused it at merge time with *"the attestation must come from the PR author (aseemann); gh is authenticated as CybotTM"*.

The correct path was already documented in `references/pull-request-workflow.md` — a real `APPROVED` review, which a non-author **can** give and the author cannot — but the whole section was framed as *"A bot-authored pull request takes the other path (#280)"*, so neither the script nor the prose covered a human-authored PR that someone else is finishing.

## Change

- **`scripts/pr-status.sh`** — `viewer{login}` is already in the GraphQL query (#319), so this adds a predicate rather than a request: `attestation_available = (author is not a bot) and (viewer is the author)`. Both advice sites branch on it. The second site — the plain `request-review` `why` at the non-quota path — recommended the attestation unconditionally and was found by the new test, not by reading. `viewer`, `viewer_is_author` and `attestation_available` join the `--json` contract.
- **`references/pull-request-workflow.md`** — the section drops its bot-only framing. A bot author is one way the attestation becomes unavailable, not the reason. Adds the symmetry that makes the alternative work: being someone other than the author is exactly what makes an approval count toward the gate, since a review by the author is excluded from it, so the two self-service routes never overlap.
- **`tests/test_pr_status_author_vs_viewer.sh`** — four cases against a stubbed `gh`.
- **`evals/evals.json`** — one eval for the "finish a colleague's PR" prompt.

## Backward compatibility

An empty `viewer` — an older `gh`, a stubbed response — keeps the previous behaviour rather than silently withdrawing the attestation advice. That case is asserted.

## Test plan

- [x] New test verified to **redden without the fix**: with `origin/main`'s `pr-status.sh` swapped in, the non-author case fails on all three of its new-behaviour assertions, while the bot case (#280) and the viewer-is-author case stay green in both directions — so the guard is wired to this defect and not to something adjacent.
- [x] Predicate confirmed against the live reproduction: `pr-status.sh -R netresearch/concourse-ci-skill 57 --json` now reports `author=aseemann viewer=CybotTM viewer_is_author=false attestation_available=false`.
- [x] Regression control: old and new script produce byte-identical output on a bot-authored open PR.
- [x] Full suite green — 20 shell suites under `tests/` plus `test_validate_git_command.py` (0 failures).
- [x] pre-commit hooks pass (shellcheck, markdownlint-cli2, check-json).

## One thing deliberately left out

The eval carries only its two positive assertions. A "must not say `--self-reviewed`" assertion would have been the first negative lookahead in `evals.json`, with no local runner here to prove it evaluates as intended — an assertion I cannot verify is decoration. That half is covered by the bash test, which is proven to fail without the fix.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01251HAVTtXzCvc14Zvhytwg)_


CybotTM:
Self-review: 561081443ea22f3fc13fb5b78da92e7dac29d1ff

PR #330, opened 2026-09-19
fix(pr-status): read the whole gate — classic protection, fork runs, failed reads

### LINKED ISSUE #275: pr-status.sh: a failed GraphQL read exits 0, and --watch loops on an unread gate
`die` inside `snapshot()` runs in a command substitution, so it kills only the subshell. A failed GraphQL read therefore does not stop the script: the single-shot path emits nothing and **exits 0**, and `--watch` loops forever on an unread gate.

## What happens

`snapshot()` is only ever called as `emit "$(snapshot)"` (line 1097) and `s=$(snapshot)` (line 1104). Both are command substitutions. `snapshot()` guards its read with `g=$(collect) || die "GraphQL query failed"` (lines 1042 and 1052), and `die` is `exit 2` (line 143) — but that `exit` terminates the subshell, not the script.

Observed against a valid GitHub PR while GraphQL was under a **secondary** rate limit (primary quota untouched, `core: 5000/5000`, `graphql: 5000/5000`):

```
$ pr-status.sh -R netresearch/php-modernization-skill 108
pr-status: GraphQL query failed
EXIT: 0
```

Exit 0 with no snapshot. Any caller that keys on the exit status reads a failed gate read as a successful one.

`--watch` is worse, because it does not stop:

```
pr-status: GraphQL query failed
pr-status.sh: line 1114: [: : integer expression expected
waiting:
```

With `$s` empty, `pending_required=$(jq -r '.checks.pending_required|length' <<<"$s")` yields the empty string, and line 1114's `[ "$pending_required" -gt 0 ]` throws. The loop then prints `waiting:` with an empty reason and sleeps. Each tick issues another GraphQL call, which keeps the secondary limit alive — the watch sustains the very condition it cannot read past. Left alone it heartbeats to its timeout, and `waiting:` is indistinguishable from a gate that is legitimately not open yet.

## Reproduction without a rate limit

```bash
#!/usr/bin/env bash
die() { printf 'probe: %s\n' "$1" >&2; exit 2; }
snapshot() { false || die "read failed"; echo '{"ok":true}'; }
s=$(snapshot)
echo "reached, s='${s}'"
n=$(jq -r '.checks.pending|length' <<<"$s" 2>/dev/null)
if [ "$n" -gt 0 ]; then echo "branch taken"; fi
```

```
probe: read failed
reached, s=''
probe.sh: line 7: [: : integer expression expected
EXIT: 0
```

## Relation to #250

The comment at line 193-197 already names this exact degradation — *"dies deep in the query as a bare `GraphQL query failed` — and `--watch` heartbeats into its timeout on a query that can never succeed"*. #250 fixed it by rejecting a GitLab repo spec up front, which removes one cause. The failure shape itself is unguarded: a transient GraphQL error on a perfectly valid GitHub repo reaches the same state, and a secondary rate limit is the common way to get there.

## Suggested fix

Make the read failure reach the caller rather than the subshell. Have `snapshot()` return non-zero instead of calling `die`, and check it at both call sites — `emit "$(snapshot)" || die ...` does not work either, since the substitution's status is what `emit` receives. Capture first, then test:

```bash
s=$(snapshot) || die "GraphQL query failed"
```

In the watch loop, a failed read should be reported as a failed read and either retried with backoff or given up on with a non-zero exit — never folded into `waiting:`. Guarding line 1114 with a default (`${pending_required:-0}`) silences the symptom while leaving the gate unread, so it is not sufficient on its own.

## Why it matters here

`pr-merge.sh` refuses to merge when the gate is shut, which is correct. But both scripts derive that verdict from a snapshot that can be silently empty. An empty read currently presents as "not actionable yet" rather than "not read" — the two need to be distinguishable, because only one of them is safe to wait out.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_015Mt4fn33aQDnBw6DhoiJhf)_

### LINKED ISSUE #310: pr-status.sh: fork workflow runs in action_required are invisible, so the checks rung passes with required contexts unreported
## What happened

On [netresearch/simple-ldap-go#227](https://github.com/netresearch/simple-ldap-go/pull/227) (a fork PR) I rebased and force-pushed a new head. Every workflow triggered by `pull_request` came back as `conclusion: action_required` — GitHub requires a maintainer to approve workflow runs from a fork after **every** new SHA, not just the first time.

`pr-status.sh` reported this state as:

```
checks      : 5 pass, 0 fail, 0 pending, 1 skip (of 6)
NEXT: request-review — no review on the current head (11717139)
```

`--watch` sat in the same state and never mentioned the eight runs awaiting approval.

## Why it is misleading

`0 pending` and a `NEXT:` line pointing at the review gate read as "CI is done, only the review is missing". In fact eight of the ten required contexts had not started and never would have without `gh api -X POST .../actions/runs/<id>/approve`. Before the push the PR had 29 check runs; after it, 6. The gate was not waiting on a review, it was waiting on me.

The failure mode is the one the tool exists to prevent: a `NEXT:` line that names an action which cannot unblock the PR. Acting on it (posting a review) would have produced an approved PR with most required contexts unreported.

## Suggestion

When the head is cross-repository, query `GET /repos/{owner}/{repo}/actions/runs?head_sha=<head>` and count `conclusion == "action_required"`. If any, that outranks the review rung:

```
NEXT: approve-workflows — 8 fork workflow runs await maintainer approval
  cmd : for id in $(gh api "repos/OWNER/REPO/actions/runs?head_sha=SHA" --jq '.workflow_runs[] | select(.conclusion=="action_required") | .id'); do gh api -X POST "repos/OWNER/REPO/actions/runs/$id/approve"; done
```

A weaker variant that still helps: compare the number of reported check runs against the required-contexts list from the ruleset, and refuse to call the checks rung satisfied while a required context has not reported at all. That also covers the plain "workflows have not registered yet" case, where a check immediately after a push shows `0 pending` simply because nothing has been queued.

## Environment

`pr-status.sh --version` from the installed copy at `~/.claude/skills/git-workflow` (marketplace cache 1.32.4), observed 2026-09-17.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01BuTVwhuqLiavZHQvzKcgAo)_

### LINKED ISSUE #329: pr-status.sh misses required checks that come from classic branch protection
`pr-status.sh` reports `required_contexts: []` and `undispatched: []` on a repository whose required checks come from **classic branch protection** rather than a ruleset. The pull request then shows `mergeState=BLOCKED` with no failing check, no unresolved thread and a review present, and `next.action` falls through to `investigate` with the message "check branch protection manually" — which is exactly the question the script exists to answer.

## Measured

`netresearch/orocommerce-skill#20`, a fork pull request. `pr-status.sh --json`:

```json
{
  "mergeState": "BLOCKED",
  "required_contexts": [],
  "rules_fetched": true,
  "undispatched": [],
  "next": { "action": "investigate", "why": "mergeState=BLOCKED with no failing check, no open thread and no missing review — check branch protection manually" }
}
```

`GET /repos/netresearch/orocommerce-skill/rules/branches/main` returns `[]` — there is no ruleset. `GET /repos/netresearch/orocommerce-skill/branches/main/protection` returns:

```json
{
  "required_status_checks": {
    "strict": true,
    "contexts": ["Analyze (actions)", "validate / Skill Validation", "eval / Eval Validation", "harness / Verify Harness Consistency", "DCO"]
  },
  "required_pull_request_reviews": { "dismiss_stale_reviews": true, "required_approving_review_count": 0 },
  "required_conversation_resolution": true
}
```

Four of those five contexts had not been dispatched at all: the three workflow runs sat in `action_required` (fork pull requests need per-push approval), and `Analyze (actions)` comes from CodeQL default setup, which does not run on pull requests from forks at all. So the real answer was "4 required contexts undispatched, one of them structurally impossible here", and the script had every input it needed to say so.

`rules_fetched: true` while `classic_protection` is populated is the tell: the rulesets call succeeded and returned nothing, and the classic-protection answer is never consulted for the required set.

## What closing this needs

`required_contexts` and `undispatched` to fall back to `branches/{branch}/protection` when the rulesets call returns no required-checks rule, and `next.why` to name the missing contexts instead of sending the reader to the branch settings page. `required_signatures` deserves the same treatment — that endpoint is separate again, and it was the other half of why #20 could not merge.

A repository can carry both, so the two sets should be unioned rather than one replacing the other.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_

CybotTM:
Merging this makes `pr-status.sh` report three states it previously read past: required checks that come from classic branch protection, fork workflow runs waiting for approval, and its own failed GraphQL read. Only `pr-status.sh` changes, plus three new test files; no other script and no reference text.

Each of the three is a case where the tool answered confidently about a gate it had not read — the one direction it must not get wrong.

## Required contexts came from one of the two sources

`$required` was built from `repos/:r/rules/branches/:b` alone. A repository whose required checks live in classic branch protection gets `[]` from that endpoint, so `required_contexts` and `undispatched` were empty, the checks rung saw nothing outstanding, and the ladder fell through to `investigate — check branch protection manually`. That is the question the tool exists to answer.

They are now the union of both sources, deduplicated. `$prot` is already fetched whenever the pull request is `BLOCKED` or a review is demanded, which is exactly the state an unreported required context produces — so the classic half is present whenever it decides anything, and a green same-repository pull request keeps its existing call count.

Measured on [orocommerce-skill#20](https://github.com/netresearch/orocommerce-skill/pull/20), which is the fixture: no ruleset, five required contexts from classic protection, one of them reporting.

## Fork runs awaiting approval were invisible

GitHub demands a maintainer's approval for a fork pull request's workflow runs after **every** new SHA, and an unapproved run is not a check run — it contributes nothing to the rollup. So the counts read `0 pending`, and the ladder reached `request-review` while eight required contexts had not started and could not start. A `NEXT:` line naming an action that cannot unblock the pull request is the failure mode this tool exists to prevent.

When the head is cross-repository, `GET /repos/:r/actions/runs?head_sha=…` is now queried, the unapproved runs are reported as `awaiting_approval`, and a new `approve-workflow-runs` rung outranks every review and signature branch. Its `cmd` is the approve call per run id. A failed fetch leaves the list empty: it understates, it never invents.

## A failed read exited 0

`die` inside `snapshot()` runs in the caller's command substitution, so `emit "$(snapshot)"` printed nothing and returned success, and `--json` handed an empty string to whatever parses it. `--watch` grew its own branch for this; the single-shot path now takes the same one and exits 2.

## Tests

Three files, one per defect, all against a stubbed `gh` — no network, no repo.

Each was run against the unfixed tree first and fails there. Each mutation reddens only its own file with the two others green beside it, and every mutant was checked to parse before its result counted. The fork-approval mutant reproduces the reported symptom exactly: `NEXT: request-review` while two runs wait.

Each file also carries its negative case, so none of them would pass against a script that simply always failed: a readable gate still exits 0, a same-repository pull request does not consult the Actions endpoint, a ruleset-only repository is unchanged, and a repository with neither source reports nothing required.

One existing test caught an omission while this was being written — `test_pr_status_draft_watch.sh` requires every emitted action to be classified actionable or waiting, and `approve-workflow-runs` was not yet in `ACTIONABLE`.

Full suite: 24 shell files pass, `tests/test_validate_git_command.py` passes, `shellcheck -S error` clean. `validate-skill.sh` reports 0 errors and 20 warnings on this branch, and the same 20 on `main`.

Closes #329
Closes #310
Closes #275

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_


CybotTM:
Self-review: 169e4d1ed307ddcfa6a01e501fec09e5c2e947c2

Issue #331, opened 2026-09-19
Decide: should pr-status.sh drive GitLab merge requests, or does netresearch-gitlab own that driver?


CybotTM:
## The question

Should `pr-status.sh` speak GitLab too, or is the boundary stated in [#250](https://github.com/netresearch/git-workflow-skill/issues/250) the end state?

This needs a decision rather than a patch, which is why it is not folded into #250: the boundary there is closed and measured, and adding a second forge to this script is a different commitment.

## What option 1 would mean

`pr-status.sh` detects the host and, for GitLab, reads `merge_requests/<iid>` — `detailed_merge_status`, which unlike GitHub's `mergeStateStatus` names the reason — plus `/discussions` for unresolved threads, and emits the same `NEXT:` vocabulary. The merge and rebase steps map to `glab mr merge` and `glab mr rebase`.

The argument for it: `pr-status.sh` is the only merge-gate and watch driver in the fleet, and half the fleet's work happens on `git.netresearch.de`. Every GitLab MR in the session that produced this issue was watched by a hand-rolled `glab api` poll loop written per task — exactly the shape the skill's own `references/merge-gate-watcher.md` warns against.

The argument against: two forges in one script means two gate vocabularies to keep in step, and `netresearch-gitlab` already owns GitLab conventions — the driver could live there instead and reuse this script's ladder rather than its code.

## What would settle it

Whether GitLab MR cycles are meant to be driven by this skill at all, or by `netresearch-gitlab`. That is an ownership question about the two skills, not a technical one about the script.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_

CybotTM:
Decided in favour of option 1, with a boundary: **vendor behaviour belongs here, instance behaviour stays in the `netresearch-*` skills.** GitLab's own semantics are true of any GitLab and have no other home, because no `gitlab-workflow` skill exists; `git.netresearch.de` as a host, our CI components and our group names are a different thing and do not move.

The separable half is already out: [#334](https://github.com/netresearch/git-workflow-skill/pull/334) removes the four places where this public skill named an internal one — the `description`, the body paragraph, and both `die()` messages in `pr-status.sh`, which is the line a stranger sees when they point the script at their own GitLab.

## The mapping, before any code

[...]

PR #334, opened 2026-09-19
fix: stop routing public consumers to an internal skill

CybotTM:
Merging this stops a public skill routing public consumers to an internal one. Five places told the reader to use `netresearch-gitlab`, which lives on `git.netresearch.de` and cannot be installed from outside.

## Where they were

- the `description` — the first thing a router reads
- the "Not Here" paragraph in the body
- **both `die()` messages in `pr-status.sh`** — the worst of them, because that is the line a stranger sees the moment they point the script at their own GitLab
- `commands/pr-finish.md` — the file a consumer *runs*, which still said "at Netresearch, the `netresearch-gitlab` skill's `references/glab_cheatsheet.md`"

All five now point at `references/pull-request-workflow.md` § *GitLab: the same gate with `glab`*, which carries the `glab api` equivalent and names nothing internal.

The pointer itself needed fixing too. Every one of these — including the text that was there before this branch — named § *GitHub only*, and there is no such section: that string is a bold lead-in mid-paragraph, while the GitLab gate lives under the heading above. A reader searching for it finds nothing. The example host in `pr-finish.md` is now `host.example.com`, since any host demonstrates a rejected `-R`.

## Behaviour

Unchanged: both refusal forms still exit 2 before any network call. Run on this branch rather than quoted from memory:

```
$ pr-status.sh -R git.netresearch.de/group/project 1
pr-status: "git.netresearch.de/group/project" is not owner/repo. pr-status.sh is GitHub-only;
for a GitLab merge request see references/pull-request-workflow.md § "GitLab: the same gate with glab"
$ echo $?
2
```

`shellcheck` clean on `pr-status.sh`, `markdownlint-cli2` clean, the 26-file shell suite and the Python gates pass, `verify-harness.sh` reports Level 3 COMPLETE, and version parity holds. `validate-skill.sh` still fails on `SKILL.md is 610 words (max 500)` — pre-existing and 611 on `main`, so this branch takes one word off it rather than adding.

## The boundary this leaves

Vendor behaviour versus instance behaviour. GitLab's own semantics — `detailed_merge_status`, `/discussions` for unresolved threads, `glab mr merge` — are true of any GitLab and have no home outside this skill, because no `gitlab-workflow` skill exists. What is specific to our installation stays where it is.

Whether this skill should grow the GitLab path rather than only document it is [#331](https://github.com/netresearch/git-workflow-skill/issues/331). This PR does not pre-empt that: it removes a reference that is wrong under either answer.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NwVe7tUuFV2HrZJ8Hkr5qJ)_


CybotTM:
Self-review: 9a12d5e59ffd2a60f4eb319fce5a7c0f5595c0f2

PR #339, opened 2026-09-20
docs(advanced-git): two measured git behaviours promoted from agent memory

CybotTM:
Merging this records two git behaviours that cost data or invite a wrong fix, both measured on git 2.55.0 rather than recalled: `git worktree remove` deletes ignored files without asking, and a bare clone's missing fetch refspec makes `--force-with-lease` reject with `(stale info)`. Both sit next to the section they correct; nothing is removed.

These come from a `/retro promote` pass over accumulated local agent memory. Nineteen candidates in the git area were checked against every reference on current `main`. Most did not belong here: five describe one operator's environment and identities, one is about the agent harness rather than git, one is a SonarCloud/Pages decision, and one is already in the GitLab skill's cheatsheet. Two were verified as already covered — `advanced-git.md:1638` for never running `git checkout <ref> -- <path>` over an uncommitted change, and `advanced-git.md:1876` for staging by name instead of `git add -A`.

**The two changes, one commit each.**

- `advanced-git.md`, "Leave the worktree before you remove it" — the section credits the missing `--force` with catching an unintended removal. That guard only covers tracked files. A worktree holding an ignored `.env` and an ignored `build/out.txt` was removed by a plain `git worktree remove`: exit 0, no warning, both files gone. Those are the files that exist in no other copy, and nothing in git can recover them because they were never in the object store. Adds the `status --porcelain --ignored` read that shows what would be lost.
- `advanced-git.md`, "Bare-Repo Layouts" — `git clone --bare` leaves `remote.origin.fetch` empty, so the clone creates no `refs/remotes/*` and the lease has no recorded remote state to compare against. The push is rejected with `(stale info)`, which reads as "somebody else pushed" and invites plain `--force`, dropping the protection entirely. Records both remedies, each measured: passing the lease its value from `git ls-remote`, and repairing the clone once by setting the refspec and fetching. The second is what a long-lived bare layout wants — with no refspec, `git fetch origin` updates nothing, so every later "is my branch behind?" question is answered from refs that never move.

**On where the second one sits.** The note this came from blamed `--depth 1`, and that is wrong: a shallow clone does carry `refs/remotes/origin/*` and pushes with an ordinary `--force-with-lease` without complaint. Measured both before writing, which is why this landed under Bare-Repo Layouts rather than under Shallow Fetches.

**Verification.** `validate-skill.sh` reports 0 errors and 20 warnings both on this branch and on unmodified `main` — identical.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NxSeVq1hDnGBCqKLGcjQ6m)_

https://claude.ai/code/session_01NxSeVq1hDnGBCqKLGcjQ6m


CybotTM:
## Review of this PR

Copilot was unavailable — the account's review quota is exhausted, so the `copilot_code_review` ruleset cannot be satisfied on any PR until it resets. Recording my own read of the diff so the decision rests on a review rather than on the gate being unsatisfiable.

**Both claims in this diff were measured before being written, not recalled.** That was the point of the exercise, and it changed one of them.

`git worktree remove` deleting ignored files: reproduced on git 2.55.0 in a throwaway repository. A worktree holding an ignored `.env` and an ignored `build/out.txt` — both shown as `!!` by `status --porcelain --ignored` — was removed by a plain `git worktree remove` with exit 0, no prompt and no `--force`, and both files were gone afterwards. The existing section credits the missing `--force` with catching unintended removals, which is true only for tracked content.

The `--force-with-lease` claim came from a note blaming `--depth 1`, and **that is wrong**. A shallow clone carries `refs/remotes/origin/main` and pushed with an ordinary `--force-with-lease` without complaint. The condition is `git clone --bare`, which leaves `remote.origin.fetch` empty so no tracking refs exist at all; that produced `! [rejected] HEAD -> featA (stale info)` on the first try. Both remedies in the text were then measured too: `--force-with-lease=<branch>:<sha>` from `git ls-remote` pushed successfully, and setting the refspec plus one `git fetch` made the ordinary form work. This is why the section landed under Bare-Repo Layouts rather than under Shallow Fetches.

**Scope.** One file, 44 added lines, nothing removed. No `SKILL.md` change, so no version-parity or body-line-count effect. Seventeen further candidates from the same promote pass were left out — most describe one operator's environment rather than git behaviour, and two were verified as already covered (`advanced-git.md:1638` and `:1876`).

**Verification.** `validate-skill.sh` reports 0 errors and 20 warnings both on this branch and on unmodified `main`. 0 unresolved threads. Rebased onto `12b86b5`; this repository's `main` moved three times during the run, and its classic branch protection has `strict: true`, so the rebases were required rather than cosmetic.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NxSeVq1hDnGBCqKLGcjQ6m)_

CybotTM:
Self-review: 8c86328d8baa89880a9ed4976db4ed72455ba424

